PR #47, opened 2016-11-20
ArtNet Timecode using OLA

offtools:
Hi,

this is a plugin which sends ArtNet Timecode by using OLA. It needs a running OLA daemon on your computer to work. (https://www.openlighting.org/ola/)
It can be used to trigger light cues on lighting desks or software by the timecode generated from a running audio file inside Lisp. I testet it under Chamsys MagicQ.

I did to get a bit into your code. Hope its not too messy.

offtools.


FrancescoCeruti:
Looks good, I will test it ASP, anyway I've notice two things:
 * `AppSettings` is imported from `lisp.application` should be `lisp.ui.settings.app_settings`
 * this can be a `module` instead of a plugin (not a big difference, probably this distinction will disappear in future)

FrancescoCeruti:
So, I'm testing the plugin, I've noticed those major points:
 * subclassing `CueTime(s)` (`TimecodeHandler`) create some problems (problem in LiSP code, already found a fix)
 * I think `TimecodeHandler` doesn't need to be a `CueTime` subclass
 * I've notice that you keep only one `OlaTimecode` object, and it keeps only one `TimecodeHandler`, so you can actually send timecodes only from one cue at the time, and any new started cue (that use timecode) interrupt the previous one.

Is the only-one-cue behavior something planned? It must be like this for some reason? I'm no expert of this protocols so I need your help there :-)

In any case I've already done some change to reflect changes in the `master` branch, and improved a bit the UI.
If you answer the above question I can procede to change those behavior myself and push them on this pull-request branch so you can review them.

FrancescoCeruti:
> If you answer the above question I can procede to change those behavior myself and push them on this pull-request branch so you can review them.

If you want, you can allow me to do so, see [docs](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)

offtools:
Hi,

feel free to add changes. 

The per one cue behaviour is necessary, because you can broadcast only one timecode at the time. The timecode is like a global clock, you can reset it, but you should not mix two clocks. (The trick with replacing the HOUR by a static audio track number is already just a hack, normaly timecode means one continous stream - no wormholes ;) )

I've noticed a little problem. CueTime.notify emits only per 1/10 seconds. This means the timecode is sended only every second or third frame (depending on the format). For me this is ok. A higher resolution would be more nice.

nodiscc:
@offtools this looks like a very promising feature, thanks!

 If you'd like to describe briefly how it works in the [Wiki](https://github.com/FrancescoCeruti/linux-show-player/wiki/ArtNet-Timecode) (eg give an example of a working setup with MagicQ + USB/DMX interface) it would be nice. Otherwise can you give us a short howto here, and I'll edit the wiki if you don't have time.

FrancescoCeruti:
@offtools I've done some changes, and merged my master allow them to work correctly, next next-step is to add an option for higher-resolution to allow 1-frame precision, only doubt, is repeating the same code a problem? in that case I should add a check to avoid that because unfortunately I cant afford a perfect timing :cry:

@nodiscc I think he has already done some wiki-work, but I think it's better to wait for the actual release to add those changes to the wiki.

FrancescoCeruti:
there are some problems, I'll fix them ASP

offtools:
Hi,

tried your changes, great. The new clock works well, I can receive every frame now. I just noticed that use_hours was handled wrong, my variable name was misleading. I changed it to replace_hours and corrected the test in __send_timecode.

FrancescoCeruti:
Ok :+1: 

**What about your name in the license? or at least a reference, after all is your work** :wink:

I bevile that now it can be merged, let me know.

Next plans are (for a possible 0.4.3 version):
 * make an `OLA` (LiSP) module to load it (when possible) and mange some common stuff
 * make the `Timecode` plugin generic so that can support multiple protocols (e.g. MIDI) like the `Controller` plugin.

offtools:
Ideas for 0.4.3 sounds good

PR #52, opened 2016-12-02
Osc dev

offtools:
Hi,

I added module for basic OSC support. It depends on pyliblo (http://das.nasophon.de/pyliblo/)

It supports global messages:
/list/go [when using list layout]
/lisp/start 
/lisp/pause 
/lisp/stop

It can capture messages to trigger Cue Actions on Cues and can send OSC messages with OSC Cues.


FrancescoCeruti:
:+1: great, give me some time to work on some of the opened `issues` than I'll take a look at your work.

I've looked only at `cue` code, to make it work correctly with the last changes in the master you should:
 * remove the `state` method (property)
 * return False in the `__start__` method

Anyway you can take a look at the other cues I've update.

aroomthedoomed:
Hi guys

would it be possible to include an OSC cue to trig other softwares with custom OSC messages? a bit like the MIDI cue ?

FrancescoCeruti:
@aroomthedoomed  It's already included in @offtools work here :)

offtools:
Hi,

I' ve made the requested changes. No need to hurry anyway. 

I plan to add fades to the OSC cues, to have the possiblity to run fades with OSC from a given start to a given end value by using the fade time setting of the cue. I'm not sure how to change the osc gui for that. I'm using a table view now with your delegate classes. Is there a to disable (greyout) single widgets inside the view. For bool and strings fade settings will make no sense.

FrancescoCeruti:
@offtools I need to take a look at the current code/UI to be 100% sure, but probably you need to subclass the model to disable the cells/delegates depending on what kind of value the row is storing.

offtools:
Maybe it makes not much sense to subclass model at this point, the reasons were only cosmetic for the osc cue argument table. It would not improve or add new to the functionality of the Osc Cue module. I keep that idea in mind for later.

Another thing. With Osc Cue and Volume Cue there are too modules that create fades, but without using the Fader class, because they perform fades on non-Properties or members from in the same class. Maybe it makes sense to make possible, that booth modules also can use the Fader class.

FrancescoCeruti:
👍 ... about the `Fader`, yeah, the idea is that `Volume Cue` should use it, I've not yet implemented those changes, but it's planned, anyway, the `Fader` target can be any (float) attribute so you should already be able to use it.

FrancescoCeruti:
I've committed some changes into the `master` branch, now `VolumeCue` is implemented using a `Fader ` object, feel free to take a look ;)

offtools:
Nice, I will adapt OscCue to that :)

PR #54, opened 2016-12-04
volumeSlider updates on running Volume Cue

offtools:
The volume slider in CartLayout was not updated by "Volume Cues". When touching the volume fader after a volume cue fade, this caused big a jump of the audio level.


FrancescoCeruti:
There are a 2 main reason to explain why it's done this way:
 1. the `volume` property you are changing is the value that will be saved into the session-file, and the volume-slider and volume-cue should only be limited to "live" control.
 2. the slider will be updated every 0.01 second, almost a waste of CPU, considering that you probably don't use the volume-slider during those fades. 

I'm aware of the problem but it should be solved providing a complete and effective "live/runtime-properties" implementation, this is something already planned to provide generic fade controls over (supported) cues properties.

PR #56, opened 2017-01-03
Timecode Output module (midi and artnet)

offtools:
Hi, 

I made timecode to a module and added midi support. I  could only test it with software, sending timecode to chamsys, not with midi hardware. Not sure if could be a difference, I'm not very used to Midi.

Would be nice, if you could check the code. I had to add an workaround to catch ola ImportErrors in backends/__init__.py. Maybe there's a better way to do it.

You are free to change things :)


FrancescoCeruti:
Hi,

As soon as the 0.4.2 is "completed" I'll take a look, shouldn't take too long :-)

FrancescoCeruti:
So, I've given an initial look, and I have two question.
First: why did you create a standalone module instead of doing all the work inside the `timecode` plugin?

I ask mainly because I don't see any usage for the timecode module outside of the timecode plugin ... and it may be a bit confusing having it split like that.

Second: can I consider it complete? so I can start merging and doing some refactoring were needed :robot:

offtools:
Hi,

Think I mixed something up between modules and plugins. Still figuring things out. Youre right the module stuff is not needed. I guess I originally also wanted to add some kind of action_cue for timecode or I moved some stuff into modules because of the AppSettings. As you mentioned it, I found some inconsistencies beause og this. From a functional side it is complete, but it needs some cleanup. Let me move all into plugins first and than merge.

FrancescoCeruti:
@offtools no problem, anyway the modules/plugin thing will change to something more clear, likely in this release

offtools:
Hi,
I moved all the code into plugins/timecode/ and removed and renamed some classes. TimecodeOutput is obsolet and moved into TimecodeCommon.  Backend is now called Protocols (like in the controller plugin). HRes clock is now always used. I removed "enabled" and "HRes" from the Appsettings. There was also a little bug, preventing sending hours > 0, when replace_hours is off.

I tested all the changes with MIDI. 
ArtNet Timecode is not working at the moment, because of OLA is imcompatible to protobuf-3.2. I hope they fix it soon. But it should work if OLA is back.

You can start to merge it.

FrancescoCeruti:
👍 perfect

PR #57, opened 2017-01-09
Add Ctr+click selection for cue list layout

Yinameah:
As you asked, I changed for better variables names.

Now we can select cues in a cue list with Ctrl+Click !

### DIFF
--- a/lisp/layouts/list_layout/layout.py
+++ b/lisp/layouts/list_layout/layout.py
@@ -152,6 +152,7 @@ class ListLayout(QWidget, CueLayout):
         self.listView.currentItemChanged.connect(self.__current_changed)
         self.listView.context_event.connect(self.context_event)
         self.listView.key_event.connect(self.onKeyPressEvent)
+        self.listView.select_cue_event.connect(self.select_event)
         self.layout().addWidget(self.listView, 1, 0, 1, 2)
 
         # PLAYING VIEW (center right)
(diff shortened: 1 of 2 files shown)


lisp/layouts/list_layout/layout.py
```diff
@@ -290,6 +291,11 @@ def double_clicked(self, event):
         if cue is not None:
             self.edit_cue(cue)
 
+    def select_event(self, event):
+        self._context_item = self.listView.itemAt(event.pos())
```
FrancescoCeruti: Assigning the selected item to `self._context_item` it's not harmful but it shouldn't be done there.
Use a new local variable (eg: `item = ...`)

Yinameah:
Hello,

Actually I'm a bit confused with that. I copy the structure from 
`def select_event(self, event):` just after. (l. 299)

For what I understand, it's precisely assigning the '<cue_list_item object>' to that variable that allow `def select_context_cue(self)` (l.235) to toggle the 'selected state'.
And the variable is only used inside the 'ListLayout' Class, which makes sense I think.
What am I missing here ?

Yinameah:
Ok, now I get it, `_context_cue` allow `CueLayout.show_context_menu()` to work. And indeed, no need in my case. I made the changes :+1: 
Now, I'm not familiar enough with Github. Should I modify the commit, or add a new one, make a new pull request ?

Something else as we are there. At the beginning of the file, there are these lines : `QT_TRANSLATE_NOOP('LayoutDetails', 'Space to execute the current cue'),`
I didn't found where there are used but since it's explanation about interaction, shouldn't we add a line about ctrl+click behavior ?

Thanks for your time, and sorry for stumbling around a little bit, I will improve quickly...

FrancescoCeruti:
Just commit, the pull-request tack all the changes pushed to the branch :)

`NAME`, `DESCRIPTION` and `DETAILS` are used in `ui/about` to show information on the layouts, probably the need to be updated, anyway, if you want, add the additional note about the `CTRL+Click`

Yinameah:
Ok, everything done.

PR #61, opened 2017-02-05
Push the French translation to 100%

trebmuh:
FYI : I didn't build and test since last time I tried, it didn't work on my current system (Debian Jessie). That said, i've double-checked and it seems correct to my eyes. Please double-check that it builds correctly and tell me if anything goes wrong due to this pull request.

PS: I'll certainly switch to Debian Stretch (where LSP is buildable) when it'll become stable, that's why I keep updating the French translation.

### DIFF
--- a/lisp/i18n/lisp_fr.ts
+++ b/lisp/i18n/lisp_fr.ts
@@ -5,22 +5,22 @@
     <message>
         <location filename="../ui/about.py" line="53"/>
         <source>Authors</source>
-        <translation type="unfinished"></translation>
+        <translation>Auteur</translation>
     </message>
     <message>
         <location filename="../ui/about.py" line="53"/>
         <source>Contributors</source>
-        <translation type="unfinished">Contributeurs</translation>
+        <translation>Contributeurs</translation>
     </message>
     <message>
         <location filename="../ui/about.py" line="53"/>
         <source>Translators</source>
-        <translation type="unfinished"></translation>
+        <translation>Traducteurs</translation>
     </message>
     <message>
         <location filename="../ui/about.py" line="75"/>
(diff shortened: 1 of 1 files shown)


FrancescoCeruti:
Thanks :+1: 

I'm trying building an AppImage, if I'm able to build a working one you should be able to use the new version :)

FrancescoCeruti:
@trebmuh You have missed the plugins/modules files, sorry I didn't notice that, and I've already merged :sob:

nodiscc:
I suggested some changes for Fade-Out/In terminology. @trebmuh can you check my review as well? Let me know if it makes sense.

FrancescoCeruti:
@nodiscc sorry, I've merged a bit too quickly :(

trebmuh:
@FrancescoCeruti : I'm looking into the modules right now. Would it be possible to get only one ts file including all the translations from the different files? From a translators point of view, it'd make things way easier.

trebmuh:
@nodiscc : since it's already merged, I've got no access (as far a I know) to your review. Please feel free to make a PR against master or to tell me about your suggested changes here, as you want.

FrancescoCeruti:
@trebmuh mmh, that's possible, since I'm planning some changes on the modules/plugins mechanism I'll probably change that too, so that at least the modules and plugins provided by default will have the translations in the main `lisp_*.ts` files.

trebmuh:
@FrancescoCeruti : that would be awesome!

OK, I've done it the modules/plugins translations (it was only 2 of those).

Let me know how it goes.

FrancescoCeruti:
okay, I'll let you know asap, thanks again :)

PR #65, opened 2017-02-16
Regex rename cue

Yinameah:
Batch renaming cue module. Here to test easily, still need some work

### DIFF
--- a/lisp/modules/rename_cues/rename_cues.py
+++ b/lisp/modules/rename_cues/rename_cues.py
@@ -38,13 +38,18 @@ class RenameCues(Module):
         MainWindow().menuTools.addAction(self.menuAction)
 
     def rename(self):
-        # Initiate rename windows
-        renameUi = RenameUi(MainWindow())
-        # Different behaviour if no cues are selected
-        if Application().layout.get_selected_cues() == []:
-            renameUi.get_all_cues()
+        # Test if some cues are selected, else select all cues
+        if Application().layout.get_selected_cues():
+            selected_cues = Application().layout.get_selected_cues()
         else:
-            renameUi.get_selected_cues()
+            #TODO : implement dialog box if/when QSettings is implemented
+            # the dialog should inform the user that rename_module load only selected cues if needed
+            # but it will bother more than being useful if we can't provide a "Don't show again"
+            # Could be provided by QErrorMessage if QSettings is supported
+            selected_cues = Application().cue_model
+
+        # Initiate rename windows
+        renameUi = RenameUi(MainWindow(), selected_cues)
 
         renameUi.exec_()
 
--- a/lisp/modules/rename_cues/rename_cues.py
+++ b/lisp/modules/rename_cues/rename_cues.py
@@ -45,7 +45,7 @@ class RenameCues(Module):
     Name = 'RenameCues'
 
     def __init__(self):
-        self._gain_thread = None
+        #self._gain_thread = None
 
         # Entry in mainWindow menu
         self.menuAction = QAction(translate('RenameCues',
@@ -67,8 +67,8 @@ class RenameCues(Module):
             renameUi.exec_()
 
             if renameUi.result() == QDialog.Accepted:
-                cues = Application().layout.get_selected_cues()
-                cues[0].name = renameUi.previewEdit.text()
+                print('Actually modification of the cues')
+
 
     def terminate(self):
         MainWindow().menuTools.removeAction(self.menuAction)
--- a/lisp/modules/rename_cues/rename_cues.py
+++ b/lisp/modules/rename_cues/rename_cues.py
@@ -41,19 +41,24 @@ class RenameCues(Module):
         MainWindow().menuTools.addAction(self.menuAction)
 
     def rename(self):
-        # Warning if no cue is selected
+        # Initiate rename windows
+        renameUi = RenameUi(MainWindow())
+        # Different behaviour if no cues are selected
         if Application().layout.get_selected_cues() == []:
             msg = QMessageBox()
             msg.setIcon(QMessageBox.Information)
-            msg.setText('You have to select some cues to rename them')
+            msg.setText(translate('RenameCues',
+                "No cues are selected. Rename tool will display all your cues.\n"))
             msg.exec_()
-        else:
-            renameUi = RenameUi(MainWindow())
-            renameUi.get_cues_name()
-            renameUi.exec_()
 
-            if renameUi.result() == QDialog.Accepted:
-                renameUi.record_cues_name()
+            renameUi.get_all_cues()
+        else:
+            renameUi.get_selected_cues()
+
+        renameUi.exec_()
+
+        if renameUi.result() == QDialog.Accepted:
+            renameUi.record_cues_name()
 
     def terminate(self):
         MainWindow().menuTools.removeAction(self.menuAction)
--- a/lisp/modules/rename_cues/rename_cues.py
+++ b/lisp/modules/rename_cues/rename_cues.py
@@ -17,7 +17,6 @@
 # You should have received a copy of the GNU General Public License
 # along with Linux Show Player.  If not, see <http://www.gnu.org/licenses/>.
 
-import gi
 from PyQt5.QtWidgets import QAction, QDialog, QMessageBox
 
 from lisp.application import Application
@@ -25,8 +24,6 @@ from lisp.core.module import Module
 from lisp.ui.mainwindow import MainWindow
 from lisp.ui.ui_utils import translate
 from .rename_ui import RenameUi
-
-gi.require_version('Gst', '1.0')
 
 
 class RenameCues(Module):
@@ -45,12 +42,6 @@ class RenameCues(Module):
         renameUi = RenameUi(MainWindow())
         # Different behaviour if no cues are selected
         if Application().layout.get_selected_cues() == []:
-            msg = QMessageBox()
-            msg.setIcon(QMessageBox.Information)
-            msg.setText(translate('RenameCues',
-                "No cues are selected. Rename tool will display all your cues.\n"))
-            msg.exec_()
-
             renameUi.get_all_cues()
         else:
             renameUi.get_selected_cues()
--- a/lisp/modules/rename_cues/rename_cues.py
+++ b/lisp/modules/rename_cues/rename_cues.py
@@ -64,6 +64,7 @@ class RenameCues(Module):
             msg.exec_()
         else:
             renameUi = RenameUi(MainWindow())
+            renameUi.get_cues_name()
             renameUi.exec_()
 
             if renameUi.result() == QDialog.Accepted:
--- a/lisp/modules/rename_cues/rename_cues.py
+++ b/lisp/modules/rename_cues/rename_cues.py
@@ -1,0 +1,75 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Linux Show Player
+#
+# Copyright 2012-2016 Francesco Ceruti <ceppofrancy@gmail.com>
+#
+# Linux Show Player is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Linux Show Player is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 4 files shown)


lisp/modules/rename_cues/rename_cues.py
```diff
@@ -0,0 +1,64 @@
...
+#
+# Linux Show Player is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Linux Show Player.  If not, see <http://www.gnu.org/licenses/>.
+
+import gi
```
FrancescoCeruti: Not needed, maybe a copy/past remaining?

lisp/modules/rename_cues/rename_cues.py
```diff
@@ -0,0 +1,64 @@
...
+from PyQt5.QtWidgets import QAction, QDialog, QMessageBox
+
+from lisp.application import Application
+from lisp.core.module import Module
+from lisp.ui.mainwindow import MainWindow
+from lisp.ui.ui_utils import translate
+from .rename_ui import RenameUi
+
+gi.require_version('Gst', '1.0')
```
FrancescoCeruti: Related to `import gi` (not needed)

lisp/modules/rename_cues/rename_cues.py
```diff
@@ -0,0 +1,64 @@
...
+                                    'Rename Cues'), MainWindow())
+        self.menuAction.triggered.connect(self.rename)
+
+        MainWindow().menuTools.addAction(self.menuAction)
+
+    def rename(self):
+        # Initiate rename windows
+        renameUi = RenameUi(MainWindow())
+        # Different behaviour if no cues are selected
+        if Application().layout.get_selected_cues() == []:
```
FrancescoCeruti: This dialog/check is not really needed, I think the dialog it will probably only bother the user

lisp/modules/rename_cues/rename_ui.py
```diff
@@ -0,0 +1,299 @@
...
+                'cue_preview': cue.name,
+                'selected': True,
+                'regex_groups': [],
+                'id': cue.id
+                })
+
+        self.update_preview_list()
+
+    def get_all_cues(self):
+        for cue in Application().cue_model.filter():
```
FrancescoCeruti: here you can iterate directly on the model, omitting `.filter()` part

lisp/modules/rename_cues/rename_ui.py
```diff
@@ -0,0 +1,299 @@
...
+                item = self.previewList.topLevelItem(i)
+                item.setText(0, self._cues_list[i]['cue_name'])
+                item.setText(1, self._cues_list[i]['cue_preview'])
+                if self._cues_list[i]['selected']:
+                    item.setCheckState(0, Qt.Checked)
+                else:
+                    item.setCheckState(0, Qt.Unchecked)
+
+    def populate_preview_list(self):
+        if self._cues_list != []:
```
FrancescoCeruti: No need to check if the list is empty (that can be done easily using `if <list>:`)

lisp/modules/rename_cues/rename_ui.py
```diff
@@ -0,0 +1,299 @@
...
+
+        self.update_preview_list()
+
+    def update_preview_list(self):
+        if self.previewList.topLevelItemCount() == 0:
+            self.populate_preview_list()
+        else:
+            for i in range(self.previewList.topLevelItemCount()):
+                item = self.previewList.topLevelItem(i)
+                item.setText(0, self._cues_list[i]['cue_name'])
```
FrancescoCeruti: The original name isn't changed, it shouldn't be necessary to set it's value again

lisp/modules/rename_cues/rename_ui.py
```diff
@@ -0,0 +1,299 @@
...
+                item.setText(1, self._cues_list[i]['cue_preview'])
+                if self._cues_list[i]['selected']:
```
FrancescoCeruti: Why not using the default list section (highlighted items)? It seems more intuitive and flexible (multi-selection, etc) than using a checkbox

lisp/modules/rename_cues/rename_ui.py
```diff
@@ -0,0 +1,299 @@
...
+        else:
+            for cue in self._cues_list:
+                out_string = out_pattern
+                for n in range(len(cue['regex_groups'])):
+                    pattern = f"\${n}"
+                    try:
+                        out_string = re.sub(pattern,
+                                            cue['regex_groups'][n], out_string)
+                    except IndexError:
+                        logging.info("Regex error : Catch with () before display with $n")
```
FrancescoCeruti: no need to log, you will fill the debug output with non-useful information

lisp/modules/rename_cues/rename_ui.py
```diff
@@ -0,0 +1,299 @@
...
+        self.layout().addWidget(self.resetButton, 5, 1)
+
+        # Spacer
+        self.layout().addItem(QSpacerItem(30, 0), 3, 2)
+
+        # Regex lines
+        self.outRegexLine = QLineEdit()
+        self.outRegexLine.setPlaceholderText(
+            translate('RenameCues', 'Rename all cue. () in regex below usable with $0, $1 ...'))
+        self.outRegexLine.setToolTip(
```
FrancescoCeruti: maybe use a small help button (with a "?" icon) to display an dialog with those info. The tooltip will hardy display all this text

lisp/modules/rename_cues/rename_ui.py
```diff
@@ -0,0 +1,299 @@
...
+            translate('RenameCues', 'What you type here is used to rename all cues.\n'
+                                    'The regex line below is used to match patterns, and to '
+                                    'capture parts of the text with parenthesis.\n'
+                                    'You can use the content of first () here with $0, the second with $1, etc..'))
+        self.outRegexLine.textChanged.connect(self.onOutRegexChanged)
+        self.layout().addWidget(self.outRegexLine, 3, 3)
+
+        self.regexLine = QLineEdit()
+        self.regexLine.setPlaceholderText(translate('RenameCues', 'Type your regex here :'))
+        self.regexLine.setToolTip(translate('RenameCues', 'Match patterns with standard python regex.\n'
```
FrancescoCeruti: same as the above comment, you can use one "help" dialog for everything.

lisp/modules/rename_cues/rename_ui.py
```diff
@@ -0,0 +1,299 @@
...
+    def onUpperButtonClicked(self):
+        for cue in self._cues_list:
+            if cue['selected']:
+                cue['cue_preview'] = cue['cue_preview'].upper()
+        self.update_preview_list()
+
+    def onRemoveNumButtonClicked(self):
+        regex = re.compile('^[^a-zA-Z]+(.+)')
+
+        def remove_numb(input):
```
FrancescoCeruti: why defining this function? it seems an overkill, simply put this code in the for loop

lisp/modules/rename_cues/rename_ui.py
```diff
@@ -0,0 +1,299 @@
...
+                'cue_preview': cue.name,
+                'selected': True,
+                'regex_groups': [],
+                'id': cue.id
+            })
+
+        self.update_preview_list()
+
+    def update_preview_list(self):
+        if self.previewList.topLevelItemCount() == 0:
```
FrancescoCeruti: Why not simply using the `populate_preview_list` when needed instead of having this check?

lisp/modules/rename_cues/rename_ui.py
```diff
@@ -0,0 +1,299 @@
...
+        # OK / Cancel buttons
+        self.dialogButtons = QDialogButtonBox()
+        self.dialogButtons.setStandardButtons(QDialogButtonBox.Ok |
+                                              QDialogButtonBox.Cancel)
+        self.layout().addWidget(self.dialogButtons, 6, 3)
+
+        self.dialogButtons.accepted.connect(self.accept)
+        self.dialogButtons.rejected.connect(self.reject)
+
+    def get_selected_cues(self):
```
FrancescoCeruti: this name is confusing, usually `get` prefix are used for functions that return something from the object, you can use `load`, or move the logic of this function in the `__init__` so that is executed only one time.

Same thing for the function below "`get_all_cues`"

lisp/modules/rename_cues/rename_cues.py
```diff
@@ -0,0 +1,64 @@
...
+            msg = QMessageBox()
```
FrancescoCeruti: This dialog/check is not really needed, I think the dialog it will probably only bother the user

lisp/modules/rename_cues/rename_cues.py
Yinameah: Yes, some copy glitch. removed.

lisp/modules/rename_cues/rename_cues.py
Yinameah: same here

lisp/modules/rename_cues/rename_ui.py
Yinameah: I agree. Done.

lisp/modules/rename_cues/rename_ui.py
Yinameah: Same also. Done. :+1:

lisp/modules/rename_cues/rename_cues.py
Yinameah: I made that because I hate to find a shortcut in a software after having used the long method for some time. And with "auto-selection", a user can use the module for ages before realizing that only selected cues are loaded in the module.

But I agree with you, It will be more annoying that useful. It's the king of dialog who should appear only once as a tip, and never after.
But I suspect there is no available "framework" for such tooltip. I removed it for now. We will see one day.

FrancescoCeruti: Okay, I see your point here. Qt provide a "Don't show again" with [QErrorMessage](http://doc.qt.io/qt-5/qerrormessage.html#details) but then you need to save them using QSettings, in future I may support this properly.

lisp/modules/rename_cues/rename_ui.py
Yinameah: Ok, very cool

lisp/modules/rename_cues/rename_ui.py
Yinameah: I didn't knew that. Done

lisp/modules/rename_cues/rename_ui.py
Yinameah: Ok. I thought an error should never pass silently.... Are we sure that no other IndexError is possible here ? Or should we put it in debug level ?

lisp/modules/rename_cues/rename_ui.py
Yinameah: Yes, indeed, seem to be some remaining of my trials. That logic is much better in __init__

Yinameah: Actually, I just realized that I can't call lisp.Application() in rename_ui.py without breaking my code to "Test only the UI quickly", so I'll try to find some sort of middle path....

lisp/modules/rename_cues/rename_ui.py
FrancescoCeruti: Mmmh, if you think it's worth logging it use the debug, but in this case it's more something that you will usually handle with a if/else than a proper "error" but try/except it's okay, see [here](https://docs.python.org/3/glossary.html?highlight=eafp#term-eafp)

lisp/modules/rename_cues/rename_ui.py
Yinameah: Yes, this test is not needed anymore

lisp/modules/rename_cues/rename_ui.py
Yinameah: True enough

lisp/modules/rename_cues/rename_ui.py
Yinameah: It seemed more complicated at the moment. But I changed it and I think it is much better now

lisp/modules/rename_cues/rename_ui.py
Yinameah: Yes indeed. It actually become even more simple without

Yinameah:
I've pushed the last changes. I think I correct and cleaned everything you reviewed.

The module seem quite usable and clean in my opinion.
Thanks for the feedbacks

FrancescoCeruti:
:+1: I'll test them. 

Last thing, you should put your name in the copyright headers in the files :wink:

PR #79, opened 2017-04-17
Cue Settings Panel

Yinameah:
Rework of my cue setting panel.
CueSettingsWidgets are re-instanciated each time so it's closer to current implementation.

Try to work with an apply button since auto record is really more complicated and need lots of modification on settings pages themselves.

### DIFF
--- a/lisp/application.py
+++ b/lisp/application.py
@@ -57,7 +57,9 @@ class Application(metaclass=Singleton):
         AppSettings.register_settings_widget(CueAppSettings)
 
         # Show the mainWindow maximized
-        self._mainWindow.showMaximized()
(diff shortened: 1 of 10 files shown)


Yinameah:
I've done some serious progress.
It's really usable now, let's call it Beta version.

The use of an Apply button allowed me to be very close of the Dialog Setting mechanism and both work alongside.

Better usability could certainly be achieved, but at the cost of rewriting a lot the settings pages themselves to be more automatically linked to properties. For the time being, it is a good middle path how already enhance the workflow speed by a long shot.

Anyone are welcome to test tracking for bugs...
capture du 2017-04-20 12-36-34

Issue #58, opened 2017-01-11
Global hotkeys/MIDI/etc...


aroom:
hi,
just a feature request. or maybe I just didn't see where to configure it. but it seems impossible yet to assign a midi message (note for example) to GO and other global function (Stop, Pause, Play, Next Cue, Previous Cue)

cheers

FrancescoCeruti:
Currently there isn't such options.

This should be quite "simple" to provide as a functionality of the `controller` plugin, this should allow to configure keyboard/MIDI/... to trigger global functions.

PR #80, opened 2017-04-24
Midi listlayout control

Yinameah:
With #58 in mind, I've worked on a plugin who allow to trigger the controls of the Cue List Layout (Go, pause, resume, etc...) with midi trigger.

Works well, but only for `note_on` message for the moment. I will try to make it work also with `control` message.
I also plan to permit triggering of "next_cue" / "previous_cue".

I already open the merge request if someone wants to test. Feedbacks are welcome.

### DIFF
--- a/lisp/default.cfg
+++ b/lisp/default.cfg
@@ -82,3 +82,5 @@ FadeInMidiMapping = No MIDI mapping
 FadeOutMidiMapping = No MIDI mapping
 ResumeMidiMapping = No MIDI mapping
 InterruptMidiMapping = No MIDI mapping
+PrevCueMidiMapping = No MIDI mapping
+NextCueMidiMapping = No MIDI mapping
--- a/lisp/default.cfg
+++ b/lisp/default.cfg
@@ -81,3 +81,4 @@ PauseMidiMapping = No MIDI mapping
 FadeInMidiMapping = No MIDI mapping
 FadeOutMidiMapping = No MIDI mapping
 ResumeMidiMapping = No MIDI mapping
+InterruptMidiMapping = No MIDI mapping
--- a/lisp/default.cfg
+++ b/lisp/default.cfg
@@ -2,7 +2,7 @@
 
 [Version]
 #Don't change this section values
-Number = 20
+Number = 21
 
 [Cue]
 FadeActionDuration = 3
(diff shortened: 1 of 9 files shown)


Yinameah:
Actually, I just test with some other controller, and my implementation work already well with `control_change` midi message, which is cool.

So, I guess, as I added `next_cue` and `previous_cue` control, the plugin is operational.

You're free to test !

FrancescoCeruti:
Okay, if I understand correctly this should be something like the global-controller that @offtools was already experimenting, but only with MIDI support, right?

As soon as I can, I'll take good look at this and your other pull-request.

Yinameah:
Hey, actually, I'm not sure what @offtools worked on exactly. I took a look on dev branch but saw nothing. (didn't look very intensely)

Yes, my implementation is very simple and straightforward only for midi. But it could be extended to other protocols. I will probably try to work on at some time.

Yes, take a look when you have time. I think you will have some feedback work, as I made some implementation choices more by trial and error. I hope there is no aberrations but I think it shouldn't be that bad.

FrancescoCeruti:
@Yinameah here https://github.com/offtools/linux-show-player/tree/controller-new/lisp/plugins/controller :)

Yinameah:
Ok, I've took a look, It seems a much deeper work than mine. And not in working state for what I can see.

I focus only on midi trigger for main ListLayout functions without changing anything else. It is maybe not as good in terms of future extendability but works and can be easily merged, or even removed once ...
I don't know, I'm not experimented enough to have a actual opinion :-). I let you and maybe @offtools have a look and say me what you think about that.

All I know is that midi trigger of `GO`, `next_cue` and `prev_cue` is really cool. Most midi controller have "play", "next" and "last" button. They are often better than keyboard keys, and LiSP wil react even if the windows loose focus, which is very useful if you go on facebook during the show (which, of course, I never did nor intend to do once ;-)

Yinameah:
So, this feature is planned to be implemented in a cleaner way in future release, probably not the way I did it here.
But since I can be useful in the mean time, and I've been asked for, I give a few instruction here in the hope it can be useful for other people how are note programmers.

WARNING : the following procedure does not install the software in a proper way. It's only having the python files in a personal directory and launch them from there. It should be fine alongside with a normal installation or AppImage, but there is no garantie.

First, you will need to have git. In case you're on Debian-like : 
`sudo apt-get install git`

Then, go into a personal folder of yours, open a terminal there, and type : 
#Get the source files of the branch with the wanted feature
`
 git clone https://github.com/Yinameah/linux-show-player.git -b midi_listlayout_control ./custom_LiSP
`
#Go to the newly created the folder
`cd custom_LiSP
`
#Switch to the origin/master branch (official lastest version) to get the latest changes
`git checkout -b master origin/master
`
#Merge midi feature and give custom name
`git merge midi_listlayout_control -m custom_merge
`


Then, all you should have to do is to launch the executable with `./linux-show-player` 
But you will have to install the dependency first. For Debian-like : 
`sudo apt-get install python3-pyqt5 python3-sortedcontainers python3-mido`

Then you can of course launch the software by double-clicking on the executable in your file manager or make a custom launcher on your desktop. In that case, give a name to it that is not ambiguous.

PR #81, opened 2017-04-25
Dialog for new Action Cue

Yinameah:
Little dialog who remember last choice to add new Cue Action
Allow rapid cue addition with only one shortcut

### DIFF
--- a/lisp/modules/action_cues/__init__.py
+++ b/lisp/modules/action_cues/__init__.py
@@ -23,7 +23,9 @@ from os import path
 from lisp.application import Application
 from lisp.core.loading import load_classes
 from lisp.core.module import Module
+from lisp.cues.cue import Cue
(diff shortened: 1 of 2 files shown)


Yinameah:
I think this is really annoying to use the menu to add a new Action Cue. But having a shortcut for each type uses too much combinaisons.

I propose this little dialog, and a unique shortcut for all Action Cue ('Ctrl+Alt+C' is just a proposition) It remember last use so it's really fast to add several cues of the same type one after another. And since it's "keyboard navigable", speed up things a lot.

Good idea ??

FrancescoCeruti:
Sorry for the (very) long delay ...
so, the idea is good, maybe we can find a better way to "register" the cues, so that all the cues, even media-cues or cues added by other modules can be listed. Right now I've a few ideas floating in my mind, when I've decided and implemented the solution we can work on this :)

PR #95, opened 2018-01-07
Set the window icon.

s0600204:
Patch to show the application icon in the window ui and desktop taskbar.

Before and After:
Before and After


FrancescoCeruti:
Thanks for the change :)
I've adjusted it to not use an hard-coded path to get the icon

PR #124, opened 2018-06-28
Layout settings fixes

s0600204:
Summary:
1. ~Provide missing default value~ Remove mention of defunct option from UI;
2. Resolve name conflict with variable `layout` in `application.py`;
3. Find the correct choice in the settings page.

(See commit messages for details)

### DIFF
--- a/lisp/application.py
+++ b/lisp/application.py
@@ -106,10 +106,10 @@ class Application(metaclass=Singleton):
         if exists(session_file):
             self._load_from_file(session_file)
         else:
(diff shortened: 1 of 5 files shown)


FrancescoCeruti:
Hi, thanks for the help!

About the missing "autoAddPage", the bug is actually that the option is still present in the UI, in fact that option has no longer any effect. I've never notice the problem because I forgot to bump the config `_version_`  (to `1.1`).

If you can remove both, the value and the option in the UI, it would be perfect :)

PR #134, opened 2018-09-04
List Layout action button fixes

s0600204:
Three small fixes for the develop branch centred around a common theme - the action buttons as displayed in the List Layout.

Hopefully the commit messages of each commit should make it clear what each is intended to do.

### DIFF
--- a/lisp/plugins/list_layout/playing_widgets.py
+++ b/lisp/plugins/list_layout/playing_widgets.py
@@ -23,11 +23,11 @@ from PyQt5.QtWidgets import QWidget, QGridLayout, QLabel, QSizePolicy, \
     QLCDNumber, QHBoxLayout
 from lisp.cues.cue import CueAction
 
-from lisp.core.configuration import AppConfig
 from lisp.core.signal import Connection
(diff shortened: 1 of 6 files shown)


lisp/plugins/list_layout/playing_widgets.py
```diff
@@ -148,19 +148,19 @@ def _update_timers(self, time):
 
     def _pause(self):
         self.cue.pause(
-            fade=AppConfig().get('listLayout.pauseCueFade', True))
+            fade=get_plugin('ListLayout').Config.get('pauseCueFade', True))
```
FrancescoCeruti: We can get the configuration once in the `__init__`, and then `self.config.get(...)`.
Ideally we should get the configuration as an `__init__` arguments.

s0600204: Done; and - although I'm not sure it's necessarily an improvement - done.

FrancescoCeruti: Thanks.
It's should keep thing simpler, in theory (see dependency-injection)

PR #135, opened 2018-09-04
Set volume in the case of no fade-up

s0600204:
When the fade-down duration of a media cue is > 0, and either:
* The fade-up duration is 0; or
* The resume-button is set to resume (instead of resume-and-fade-up) in the preferences

Then:
* Pausing will fade-down and pause track
* Resuming again will resume playback, but not restore volume

This changeset fixes that (for the `master` branch). (The `develop` branch suffers from the same problem, so I don't know if you wish to also apply this patch to that, or something else. You're the boss :wink:)

### DIFF
--- a/lisp/cues/media_cue.py
+++ b/lisp/cues/media_cue.py
@@ -64,15 +64,14 @@ class MediaCue(Cue):
     def __start__(self, fade=False):
         if fade and self._can_fade(self.fadein_duration):
             self.__volume.current_volume = 0
+        else:
+            # If we know there won't be a fade-in: restore the volume,
+            # just in case we're resuming after a fade-out-and-pause.
+            self.__volume.current_volume = self.__volume.volume
 
         self.media.play()
         if fade:
             self._on_start_fade()
-
-        # If there was a fade-down, but there wasn't a fade-up - either
-        # because the fade-up was disabled, or because it had a duration
-        # of zero - restore the volume.
-        self.__volume.current_volume = self.__volume.volume
 
         return True
 
--- a/lisp/cues/media_cue.py
+++ b/lisp/cues/media_cue.py
@@ -69,6 +69,11 @@ class MediaCue(Cue):
         if fade:
             self._on_start_fade()
 
+        # If there was a fade-down, but there wasn't a fade-up - either
+        # because the fade-up was disabled, or because it had a duration
+        # of zero - restore the volume.
+        self.__volume.current_volume = self.__volume.volume
+
         return True
 
     def __stop__(self, fade=False):


lisp/cues/media_cue.py
```diff
@@ -69,6 +69,11 @@ def __start__(self, fade=False):
         if fade:
             self._on_start_fade()
 
+        # If there was a fade-down, but there wasn't a fade-up - either
+        # because the fade-up was disabled, or because it had a duration
+        # of zero - restore the volume.
+        self.__volume.current_volume = self.__volume.volume
```
FrancescoCeruti: This should be in an else branch of the previous `if` statement (line 69).
Since `_on_start_fade()` is asynchronous, even when called, this line will get executed immediately after, this can cause some problem that might be unnoticeable when going from `stop` to `running`, due to some delay in loading the file, you can notice it when going from `pause` to `running` (with a fade).

lisp/cues/media_cue.py
```diff
@@ -69,6 +69,11 @@ def __start__(self, fade=False):
         if fade:
             self._on_start_fade()
 
+        # If there was a fade-down, but there wasn't a fade-up - either
```
FrancescoCeruti: It should be better to keep the `fade-in` and `fade-out` terms, as in other parts of code-base. (also in the line below)

FrancescoCeruti:
> I don't know if you wish to also apply this patch to that, or something else. You're the boss :wink:

Yeah, go for it  :+1:

s0600204:
> Yeah, go for it :+1:

Uh, does this mean you want me to open a pull-request for what will be the same change, applied to the same file, in the same directory, but on a different branch?

Also, I didn't/don't know when you're next planning to merge `master` into `develop` which (after this PR is merged) would also solve the problem (and would also adhere to the principle of single-source-of-truth).

lisp/cues/media_cue.py
s0600204: Hmm, sorry. Mentally I think of volumes going up or down, rather than in or out.

Hope the new comment's better.

FrancescoCeruti:
> Also, I didn't/don't know when you're next planning to merge master into develop which (after this PR is merged) would also solve the problem (and would also adhere to the principle of single-source-of-truth).

Eheh, I hope soon. I think we can leave it only on the master, I will handle it with all the various merge-conflicts

lisp/cues/media_cue.py
FrancescoCeruti: that make sense, but, at least in my experience, `fade-in/fade-out` is what is most commonly used :-)

PR #143, opened 2018-12-06
[RFC] Split up "Action cues"

s0600204:
This is a little rough, so is presented as an RFC (Request for Comments).

As things currently stand, there are two categories of cues: "Media cues" and "Action cues". There's only one of the former, whilst a good few of the latter. It is the latter group that this PR deals with.

All the cues in this latter group are defined inside the same folder, which is represented within LiSP as a single plugin. This appears to be so that they benefit from the `__init__.py` file of that plugin/folder that initialises the various cue classes and sets them up to be usable within LiSP.

However, a couple of these cues really belong to other plugins, specifically the `midi` and `osc` cues. These two cues would be perhaps better placed with their respective plugins. This is one of the things this PR does.

However, the logic required to load the cues, as presented inside the `action_cues/__init__.py` file, is something that should only be written once, rather than duplicated across multiple plugins - ideally a plugin should have an easy, one (or two) line, way of telling LiSP that "*here* is a new cue type, please register it". This the other thing that this PR does - it moves the registration code from the `action_cues` init file into the `cue_factory` file, permitting it to be used across multiple plugins.

I have also moved the "Command cue" out of the "Action cues" into its own plugin. This means that the only classes remaining in the `action_cues` folder/plugin are those of cues that act upon other cues.

The parts I could do with comments on, are:
* The code move. It's pretty much a cut-paste, with a little alteration to allow plugins to define what submenu their new cue should be listed under. Whether the approach is the best is up for debate.
* The names of the submenus in the right-click context menu/`Edit` toolbar menu. The cues-that-act-upon-other-cues are still under "Action cues", whilst the `midi` and `osc` cues are now under "Protocol cues", and the `command` cue is under "Miscellaneous cues".

Things that could also be looked into:
* Ordering the cue submenus.


fralix:
`[Speaking as a user, I'm only familiar with the basic structure of the LiSP code]`

I like the idea, but I'd prefer not to have too many categories since there's not an overwhelming amount of cue types at the moment. 
"Protocol cues" sounds weird to me, I'd prefer something like Integration Cue or Connection ~ or External ~ or something like this. 
The command cue could be available in the proposed groups, I would avoid creating a category for it as it would feel so lonely in there ;-)

PR #148, opened 2019-01-09
Save last session path

pade:
Hi,

A (very) little evolution to store the last session path, to avoid to navigate through directories from HOME to the session file.

PS: Thanks for this very nice tool ! I use it for theatre shows, linked with QLC+ for managing lights: it works very well

### DIFF
--- a/lisp/application.py
+++ b/lisp/application.py
@@ -18,7 +18,7 @@
 import json
 import logging
 from PyQt5.QtWidgets import QDialog, qApp
-from os.path import exists
+from os.path import exists, dirname
 
 from lisp import layout
 from lisp.core.actions_handler import MainActionsHandler
@@ -26,6 +26,7 @@ from lisp.core.session import Session
 from lisp.core.signal import Signal
 from lisp.core.singleton import Singleton
 from lisp.core.util import filter_live_properties
+from lisp.core.configuration import AppConfig
 from lisp.cues.cue import Cue
 from lisp.cues.cue_factory import CueFactory
 from lisp.cues.cue_model import CueModel
@@ -165,6 +166,11 @@ class Application(metaclass=Singleton):
     def _save_to_file(self, session_file):
         """Save the current session into a file."""
         self.session.session_file = session_file
+
+        # Save session path
+        last_session_path = dirname(session_file)
+        AppConfig().set("session.last_path", last_session_path)
+        AppConfig().write()
 
         # Add the cues
         session_dict = {"cues": []}
--- a/lisp/ui/layoutselect.py
+++ b/lisp/ui/layoutselect.py
@@ -30,6 +30,7 @@ from PyQt5.QtWidgets import (
 
 from lisp import layout
 from lisp.ui.ui_utils import translate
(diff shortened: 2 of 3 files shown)


fralix:
Nice! Another idea would be to store the Path in the project instead of the config, so it opens the used project folder ...

pade:
Hi fralix,
I don't really understand what you mean by saving the path in the project: you mean saging it in the session file ?
This little modification is to keep the path of the last opened session, to easily find it when you launch another time Lisp.

fralix:
@pade yes, exactly. I'm preparing some events for months, while also opening other sessions. The ideal behaviour for me would be to store the folder within the session file so it would open the corresponding directory on my PC. 
I know what the modification does, and I'm definitely happy about this (the current behaviour is sometimes annoying). But having a per session storage of this path would be the ideal behaviour for me. 
The best approach would probably be to define a default folder (not HOME but for me for example HOME/events/) and override this path if another one has been defined in the session (HOME/events/the_show_2019). 
I know this is much more work than what you already implemented, I'm already happy about this. We can move this to another feature request.

pade:
@fralix : OK, so you want to save, into the session file, the path of your project. By this, you mean the path where you put media files used in media cues of the project ? So when opening the session file, and adding a new media, you are directly in the good directory.
Do you know that in the development branch there is now a new feature that store the last media file path used ? By this way, you have to select the good directory only the first time you add a media, after, when you add another media, you are directly in the same directory.
If you think it is not enough for your need, I could take a look to implement what you need.

fralix:
Well, the last path used would usually be the right directory already, so no need to add a setting for this. 
Sorry for the thoughts in this thread ...

lisp/application.py
```diff
@@ -166,6 +167,11 @@ def _save_to_file(self, session_file):
         """Save the current session into a file."""
         self.session.session_file = session_file
 
+        # Save session path
+        last_session_path = dirname(session_file)
```
FrancescoCeruti: No need to keep a variable for this, is used in only on place, can be inline

lisp/application.py
```diff
@@ -166,6 +167,11 @@ def _save_to_file(self, session_file):
...
+        AppConfig().set("session.last_path", last_session_path)
```
FrancescoCeruti: You can use directly `self.conf`, no need to import and `AppConfig`

lisp/application.py
```diff
@@ -166,6 +167,11 @@ def _save_to_file(self, session_file):
...
+        AppConfig().write()
```
FrancescoCeruti: As above comment `self.conf`

lisp/ui/layoutselect.py
```diff
@@ -99,8 +100,12 @@ def show_description(self):
         )
 
     def open_file(self):
+        last_session_path = AppConfig().get("session.last_path", os.getenv("HOME"))
```
FrancescoCeruti: Here it could be convenient to keep a reference of `AppConfig` instead of calling it multiple time (e.g. `conf = AppConfig()`).
This change affect also the lines 109/110.

PR #149, opened 2019-01-22
Follow XDG spec when storing user data.

s0600204:
On Linux systems, it is generally recommended to follow the [XDG specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) when determining where to store user configuration, file-level caches, and user-specific application data.

Thus, as of this PR:
* *configuration* is now stored at `$XDG_CONFIG_HOME/linux_show_player`
    (by default: `~/.config/linux_show_player`)
* *presets* are now stored under `$XDG_DATA_HOME/linux_show_player/presets`
    (by default: `~/.local/share/linux_show_player/presets`)
* *logs* are now stored under `$XDG_CACHE_HOME/linux_show_player/logs`
    (by default: `~/.cache/linux_show_player/logs`)

On the downside, this PR does add another python dependency. However, the advantage of using the `appdirs` python module (instead of hard-coding the `~/.{config|share|cache}`paths) is:
1) We follow the `$XDG_{*}_HOME` environment values, should a user have them set;
2) If the standard ever changes, then we don't have to worry about it;
3) If *lisp* is ever ported to Windows or OSX, then the module provides suitable paths for those OSes;

### DIFF
--- a/lisp/__init__.py
+++ b/lisp/__init__.py
@@ -27,7 +27,8 @@ __version__ = "0.6.0.dev0"
 # Application wide "constants"
 APP_DIR = path.dirname(__file__)
 
-USER_DIRS = AppDirs("linux_show_player", __author__)
+# The version passed follows <major>.<minor>
+USER_DIRS = AppDirs("LinuxShowPlayer", version=".".join(__version__.split(".")[0:2]))
 
 DEFAULT_APP_CONFIG = path.join(APP_DIR, "default.json")
 USER_APP_CONFIG = path.join(USER_DIRS.user_config_dir, "lisp.json")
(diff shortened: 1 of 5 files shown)


lisp/__init__.py
```diff
@@ -26,12 +27,10 @@
 # Application wide "constants"
 APP_DIR = path.dirname(__file__)
 
-USER_DIR = path.join(path.expanduser("~"), ".linux_show_player")
-
-LOGS_DIR = path.join(USER_DIR, "logs")
+USER_DIRS = AppDirs("linux_show_player", __author__)
```
FrancescoCeruti: I'll leave the author empty (default to application name) I think it's less confusing to have a directory with  my name around :wink:
Also, since we are already changing the setting directories, we may change it into "LinuxShowPlayer", I don't know why I've used underscores at first.
Also, I think, it could be nice to use the version parameter, only with \<major\>.\<minor\> (e.g 0.6)

Issue #150, opened 2019-01-25
Remember last song path


Gajenthran:
I think the title is quite self-explanatory (but I still prefer to detail). The idea is to save the last cue that we add on the app. For the moment, the app go to a default path to system "music" folder. 

Actually, I already do the necessary work to change the behaviour but I would like to know if it's an interesting idea.

FrancescoCeruti:
As mentioned in #151 this has already been implemented :-)

PR #151, opened 2019-01-25
#150: Remember last cue path

Gajenthran:
Hello, 
I add some changes to answer the Issue #150 (which is mine actually). I tried not to distort your code. 
Hoping it helped you ! 
(This was my first Pull Request, so I hope I have do in the right way. Extremely sorry if this is not the case)


FrancescoCeruti:
Hi! Happy to see you're interest in the project :-)

Unfortunately the proposed changes are already implemented in the development version (with persistence between different sessions).
I've also noticed that you've probably started from the `master` branch to than target the `develop` branch, this crates a lot of conflicts and should generally be avoided.

PR #154, opened 2019-02-02
Minor changes: Return to player if the user don't save and add shortcut to remove cue

Gajenthran:
Hello, 
1) I detect a minor problem when I want to save a file. 
Indeed, when I close the session and I don't save the session, a message box is activated (with 3 options: "Close without Saving", "Cancel" and "Save"). If I click in "Save" (it opens a widget with all my directories and files) and close by mistake the widget, I quit the application. However, I think it will be better if we close by mistake (and so we don't save anything), we should be redirected on the app.

2) Morever, I notice also that there is no shortcut to remove a cue. So I decided to put it (I hope it will be useful because it's an action that I use regularly).

Here are some pictures of the problem (for the first problem):
message-box
cancel
 
I hope I worked on the correct branch this time. To be honest, it was not intentional to fix two issues on the same pull request (save and shortcut), I would have preferred to make a pull request for each issue but I made a little mistake (I'am extremely sorry if it bothers you)

### DIFF
--- a/lisp/plugins/list_layout/layout.py
+++ b/lisp/plugins/list_layout/layout.py
@@ -251,7 +251,7 @@ class ListLayout(CueLayout):
             if QKeySequence(keys) in self._go_key_sequence:
                 event.accept()
                 self.__go_slot()
-            elif event.key() == Qt.Key_Backspace:
+            elif event.key() == Qt.Key_Delete:
                 cue = self.standby_cue()
                 if cue is not None:
                     self._remove_cue(cue)
--- a/lisp/plugins/list_layout/layout.py
+++ b/lisp/plugins/list_layout/layout.py
@@ -251,6 +251,10 @@ class ListLayout(CueLayout):
             if QKeySequence(keys) in self._go_key_sequence:
                 event.accept()
                 self.__go_slot()
+            elif event.key() == Qt.Key_Backspace:
+                cue = self.standby_cue()
+                if cue is not None:
+                    self._remove_cue(cue)
             elif event.key() == Qt.Key_Space:
                 if event.modifiers() == Qt.ShiftModifier:
                     event.accept()
--- a/lisp/layout/cue_layout.py
+++ b/lisp/layout/cue_layout.py
@@ -218,6 +218,9 @@ class CueLayout(HasProperties):
     def finalize(self):
         """Destroy all the layout elements"""
 
+    def _remove_cue(self, cue):
+        self.cue_model.remove(cue)
+
     def _remove_cues(self, cues):
         for cue in cues:
             self.cue_model.remove(cue)
--- a/lisp/ui/mainwindow.py
+++ b/lisp/ui/mainwindow.py
@@ -289,11 +289,10 @@ class MainWindow(QMainWindow, metaclass=QSingleton):
 
     def _save(self):
         if self.session.session_file == "":
-            saved = self._save_with_name()
-
-        if saved:
+            return self._save_with_name()
+        else:
             self.save_session.emit(self.session.session_file)
-        return saved
+            return True
 
     def _save_with_name(self):
         filename, ok = QFileDialog.getSaveFileName(
@@ -306,7 +305,6 @@ class MainWindow(QMainWindow, metaclass=QSingleton):
 
             self.save_session.emit(filename)
             return True
-
         else:
             return False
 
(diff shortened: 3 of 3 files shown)


fralix:
Hey, I like your recent commit, but do you really think that Backspace is a good key? By convention, I would expect the Del key to do that job and not the backspace, just like managing files in a file browser. 
I would also love to have a confirmation dialog, unless you hit shift + del.

But I'm very glad to see this problem addressed, thank you very much!

Gajenthran:
> Hey, I like your recent commit, but do you really think that Backspace is a good key? By convention, I would expect the Del key to do that job and not the backspace, just like managing files in a file browser.

Hi @fralix, I think that you're right, "Del" would be a better key shortcut than "Backspace" (sorry I work on a Mac that's why I thought of "Backspace") . Thank you for the idea !

> I would also love to have a confirmation dialog, unless you hit shift + del.

Ok! I will work on that, I think it's also a good idea.

Morever, after what you said, I thought to add a shortcut configuration (people can change the keys as they want). But before working on that, I would like to know if it's a good idea and if people are interested in this feature.

fralix:
Okay, thank you very much for receiving my feedback. 

A shortcut configuration is a nice idea as well to allow people to customize their workflow :+1:

FrancescoCeruti:
Hi,
first sorry for absence, second, nice work :smile:

> I hope I worked on the correct branch this time. To be honest, it was not intentional to fix two issues on the same pull request (save and shortcut), I would have preferred to make a pull request for each issue but I made a little mistake (I'am extremely sorry if it bothers you)

This time it's correct :wink: no problem

> Morever, after what you said, I thought to add a shortcut configuration (people can change the keys as they want). But before working on that, I would like to know if it's a good idea and if people are interested in this feature.

This probably require some "invasive" modification of the UI, but it could be very useful (we can also think to allow the user to save multiple presets), if you like you could start by opening an issue to discuss that

lisp/plugins/list_layout/layout.py
```diff
@@ -251,6 +251,10 @@ def _key_pressed(self, event):
             if QKeySequence(keys) in self._go_key_sequence:
                 event.accept()
                 self.__go_slot()
+            elif event.key() == Qt.Key_Delete:
```
FrancescoCeruti: This will use the platform/desktop default keybindings
```suggestion
            elif sequence == QKeySequence.Delete:
```

lisp/plugins/list_layout/layout.py
```diff
@@ -251,6 +251,10 @@ def _key_pressed(self, event):
...
+                cue = self.standby_cue()
```
FrancescoCeruti: This way you can remove multiple selected cues
```suggestion
               self._remove_cues(self.selected_cues())
```

lisp/plugins/list_layout/layout.py
```diff
@@ -251,6 +251,10 @@ def _key_pressed(self, event):
...
+                if cue is not None:
```
FrancescoCeruti:
```suggestion
```

lisp/plugins/list_layout/layout.py
```diff
@@ -251,6 +251,10 @@ def _key_pressed(self, event):
...
+                    self._remove_cue(cue)
```
FrancescoCeruti:
```suggestion
```

lisp/plugins/list_layout/layout.py
```diff
@@ -251,6 +251,10 @@ def _key_pressed(self, event):
             if QKeySequence(keys) in self._go_key_sequence:
```
FrancescoCeruti:
```suggestion
            sequence = QKeySequence(keys)
            if sequence in self._go_key_sequence:
```

Gajenthran:
> I'd suggest this changes :)
> 
> Also I don't think we need a confirmation dialog on this, we can always undo (`Ctlr+Z`)

Thank you for the suggestion. Indeed, I didn't think about removing several cues : it's a great idea !

> This probably require some "invasive" modification of the UI, but it could be very useful (we can also think to allow the user to save multiple presets), if you like you could start by opening an issue to discuss that

Yes, you're right, I think I will open an issue. For now, i'm working on another improvement (add live mode which will help us to lock/unlock certain features. For example, lock editing, removing...), I will start to open an issue to know the opinion on this feature. Hope it won't bothers you to post to many issues...

FrancescoCeruti:
> Hope it won't bothers you to post to many issues...

Don't worry :wink:

Issue #158, opened 2019-02-26
New features for Preferences 


Gajenthran:
Hello,
Here is some new features to add in Preferences (I placed them in order of importance (in my opinion) : 
- [ ] Change language
- [ ] Add a white theme (there is already a dark theme)
- [ ] Key bindings (configuration to add new shortcuts)
- [ ] Modify the size of font (small/large)

I don't think if there is an option to change the language but I would like to know how we can change the language otherwise.

FrancescoCeruti:
* **Change language** -> :+1:  Right now is possible via a command line argument
* **Add a light/white theme** -> :+1:  It was there before, but it was simply one the default qt styles, it needs some adjustment or a few widgets will look bad or not display properly
* **Key bindings** -> :+1:  as said before it require some effort
* **Modify the size of font** -> :confused: In my opinion we should rely on the OS settings, which QT already does, unless there is some convincing reason not to

Gajenthran:
> * **Change language** -> 👍  Right now is possible via a command line argument
Can I know the command, please ? Because I tried to search arguments for the language but I didn't find it.
Thanks for the reply.

FrancescoCeruti:
Yes, sorry :smile: 

It's actually the `--locale` option, for example `linux-show-player --locale it` will use the italian translations.
:question:  why *locale*? https://en.wikipedia.org/wiki/Locale_(computer_software)

Consider that beside the source language (english -> `en`) in the develop version there are many untranslated strings, if you're interested you can se at https://crowdin.com/project/linux-show-player the current situation

Gajenthran:
> Yes, sorry 😄
> 
> It's actually the `--locale` option, for example `linux-show-player --locale it` will use the italian translations.
> ❓ why _locale_? https://en.wikipedia.org/wiki/Locale_(computer_software)

Thanks for the command and the explanation ! 

> Consider that beside the source language (english -> `en`) in the develop version there are many untranslated strings, if you're interested you can se at https://crowdin.com/project/linux-show-player the current situation

Yes, I wanted to ask you about that because I saw that you modify i18n folder and there is some "unfinished" string. I will surely do that (especially french translation) !

FrancescoCeruti:
> Yes, I wanted to ask you about that because I saw that you modify i18n folder and there is some "unfinished" string. I will surely do that (especially french translation) !

[...]

PR #165, opened 2019-03-20
Choose the language in Preferences

Gajenthran:
Concerning the issue #158, I put the option to change the language.
I make sure that the --locale option on the command line still work. I put it in "General" below the "Icons theme" and "UI theme".
However, I think there is a little problem (UI problem): for the language in the combobox, I put only the 2 first letters of the language. For example, for "French", I put "fr". I don't know if it's really annoying for 9 languages but it might be annoying if we decide to add more languages.

PS: I only touch general.py, main.py, default.json and lisp.json. Extremely sorry, I create another pull request for the same issue. That's my first time so I forgot to use the command to pull.


FrancescoCeruti:
Hi :)

As a general rule, when you develop a new feature to merge in one of the upstream breaches (`develop` in this casa) you should start from a fresh copy of that branch (or a branch which is in sync with the upstream). Especially you shouldn't start from a branch where you have already other pull requests, this keep things cleaner ;)

As for the code, we can improve a few things:
 - the default locale should not be `en` but `""` or `null`, which should be interpreted as "system default", also provide this as an option in the settings combobox.
- "locale" is more suited than as a key "language", is consistent with the current terminology
- in the UI we should display the language full name, I believe that the `QLocale` class contains something that can be used, not sure. The fallback option is to build a dictionary to map "locale code" -> "full name"
- remember to increase the configuration version or the new settings might get ignored

Gajenthran:
> As a general rule, when you develop a new feature to merge in one of the upstream breaches (`develop` in this case) you should start from a fresh copy of that branch (or a branch which is in sync with the upstream). Especially you shouldn't start from a branch where you have already other pull requests, this keep things cleaner ;)

Thanks for the advice ! I follow what you said but the only thing that I didn't know is that we have to sync our works with the upstream (I thought that a simple ```git pull```was enough).

> As for the code, we can improve a few things:
> 
> * the default locale should not be `en` but `""` or `null`, which should be interpreted as "system default", also provide this as an option in the settings combobox.
Done !

> * "locale" is more suited than "language" as settings key, is consistent with the current terminology
Done ! And I completely agree with you.

> * in the UI we should display the language full name, I believe that the `QLocale` class contains something that can be used, not sure, another option is to build a dictionary to map "locale code" -> "full name"

I create a dictionary as you said, but I put it in general.py. I hesitate to create put the dictionary into a file but I prefer to let you handle that (I don't think that general.py is the right place to put the dictionary).

fralix:
I don't know if you thought about using translated language names or if it's a good idea to do so. 

(Like to use the local language names instead of international ones, German gets translated into Deutsch, French into Français etc)

Gajenthran:
Yes, I do that : it's easier. I think it's not necessary to translate into each languages.

FrancescoCeruti:
@Gajenthran I've look into the [`QLocale`](https://doc.qt.io/qt-5/qlocale.html) docs, we could use something like this:
```
ql = QLocale(locale)
combobox_text = "{} ({})".format(
    QLocale.languageToString(ql.language()),
    ql.nativeLanguageName()
)
```

For `QLocale("it")` this will output `"Italian (italiano)"`. This way we can can cover both cases :cake: 

Anyway I'll suggest to have a "custom" widget to keep things cleaner, something like `FadeComboBox` in `lisp/ui/widgets/fades.py`

Gajenthran:
Thanks for your suggestions @FrancescoCeruti !
As you said, I put ```QLocale``` method to have the name in both language.

And I create also a "custom" widget (it's clearly better because it avoids mixing ```Locale``` functionnalities in ```general.py```).

FrancescoCeruti:
I've merged it manually, with a few changes, see the commit reference above.

Thanks :smile:

Issue #160, opened 2019-03-03
Request: Cue Shortcut Key <esc>


Mr-Si:
Hi there,

I have been trying to use the "Escape" button on the keyboard as a shortcut command key for interrupting one or all cues in the form of a panic button, a bit like QLab does by default, but I have been unsuccessful so far in getting the software to recognise the key when I'm trying to create it.

Is there a way you can instruct me in how to do this? or is <esc> not a key that is allowed to be used in LiSP?

Many thanks,
Simon

FrancescoCeruti:
Hi Simon,

unfortunately right now the only accepted keyboard shortcuts (for controlling the cues) are those made of "typeable" characters, so mostly letters and numbers. I'll try to allow special characters or key sequences for next release.

Mr-Si:
Thank you Francesco, I appreciate that greatly!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Gajenthran:
I see that this feature was added to the v0.6 milestone, so I would like if it's better that I let you work on that, @FrancescoCeruti ? 

For this feature, my idea was to add two commands: one with a combobox (with special character) and another with a textarea where we can put a letter as we can see now.

An exemple of that: 
<img width="569" alt="Capture d’écran 2019-03-20 à 16 25 29" src="https://user-images.githubusercontent.com/35834540/54697298-99f39600-4b2d-11e9-93c0-65df851049c2.png">

In this example, we can also add key sequences : if we choose Escape, there will be no second word but if we choose CTRL, we have to write a second word (I forgot to add "None" in the combobox if we only want to use letters and numbers).

FrancescoCeruti:
I was thinking to use the [QKeySequenceEdit](https://doc.qt.io/qt-5/qkeysequenceedit.html) widget, It's more flexible and should handle all the reasonable scenarios (`CTRL`, `SHIFT`, `ALT`, `CMD` (Mac), `Meta` (windows key), `F1-F12`)

Gajenthran:
Oh ! I search about QKeySequenceEdit: it will more efficient than what I said and appropriate to what we want to do (actually i'm learning PyQt with you app, I didn't know the existence of this method).

fralix:
Yeah, flexibility sounds great. The method proposed by @Gajenthran didn't really cover scenarios like only the w or ctrl + shift + w for water background :-)

FrancescoCeruti:
[...]

PR #172, opened 2019-05-05
Add shortcut for cue (issue #158)

Gajenthran:
As I said on the title, I add shortcut for cues (issues #158 and #160) with ```QKeySequenceEdit```. I have not changed a lot of things : I replace the ```QLineEdit``` with ```QKeySequenceEdit``` for the first column, to have a variety of commands.  Then I modified a little bit the ```key_pressed``` function in keyboard.py to include ```QKeySequence```.

Hope it will help you !


FrancescoCeruti:
Thanks

I've tested the solution but unfortunately the QKeySequenceEdit is a bit inadequate :
• 
The widget allow to insert multiple values, this is not supported by the plugin

• 
Some keys (or combinations) cannot be entered because are intercepted elsewhere (e.g. Esc , Tab , Enter ) to navigate the widgets (e.g. Esc close the editor)

I've already implemented a solution using a custom widget that provide an alternative way to insert the mentioned keys, I'll push it when at home.

Gajenthran:
I saw it (I had a little problem with the List Layout but I saw the issue about that).

> Some keys (or combinations) cannot be entered because are intercepted elsewhere (e.g. Esc , Tab , Enter) to navigate the widgets (e.g. Esc close the editor)

 I forgot that point. 
Don't you think we should also do for Selection Mode ? For example if I had a new shortcut with UP arrow, it will allow my shortcut and the action for the selection Mode and will potentially create conflict. 

I put some lines to avoid that kind of conflict in layout.py, you can check it if you want.

Gajenthran:
Instead of creating a new pull request, I put it in this. I add more features to shorten (editing and cloning) in the list. I felt that it was not necessary to create new pull request because it's minor changement in the code.

PR #178, opened 2019-05-12
adding light theme

ghassenghouibi:
Hello,
As I said on the title, I add a light . I did'nt change a lot of things,I got helped by your dark theme.
This was to reply for issue #158 
Hope it will help you !


FrancescoCeruti:
I've tried it, and it looks a bit too white, there should be, in my opinion, some difference in the colors between certain widgets.

I'd suggest you to start from the default Qt "Fustion" style, and use the stylesheet to customize on top of that.
Ideally you would simply ensure to load the base style in `light.py` with `qt_app.setStyle("fusion")`, before the `setStyleSheet` call, you can start with an empty qss file, and see what need to be "fixed", primarily the "custom widgets" styles (from line 632 of the current qss). The "fusion" style will also supply appropriate assets.

ghassenghouibi:
Thank you for your suggestion ,i'll work on that soon

Issue #184, opened 2019-06-20
When using multi-select, fade out options don't apply correctly


radiomike:
Multi select 2 or more media cues, change fade **out** time or curve. Should result in time and curve applied to all selected cues, but nothing gets applied to any selected cue.

Multi select 2 or more media cues, change fade **in** time or curve. Should just result in Time and curve applying to fade in, but fade out time and curve also get reset to 0 and linear respectively for all selected cues. If fade out isn't touched, that should remain the same.

However, if you multi select 2 or more media cues, change fade **in** _and_ **out**, then time and curves apply correctly to all selected cues.

PR #185, opened 2019-06-21
Fix for issue #184

radiomike:
Now checks both fadeOut check values well as fadeIn check value.


FrancescoCeruti:
Thanks :+1: 

99% copy-paste error :laughing:

PR #190, opened 2019-12-14
allow config to disable go key while playing

paulbdavis:
Adds a config option to allow disabling the Go hotkey in list layout mode while there are actively playing cues. 

The Go button itself will still work, as will any programmed triggers, etc. It only disables the go key combo as to prevent accidentally triggering the next cue.

The text on the config screen is still ugly, wanted to see if this would be accepted in general before I dig into that.

### DIFF
--- a/lisp/default.cfg
+++ b/lisp/default.cfg
@@ -55,6 +55,7 @@ ShowPlaying = True
 AutoContinue = True
 EndList = Stop
 GoKey = Space
+DisableGoKeyWhilePlaying = False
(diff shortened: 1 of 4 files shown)


FrancescoCeruti:
Looks like it could be accepted :-) the only thing is that you should work with the development branch.
I'm not going to merge new features in the current master, until the development branch is merged, and it's a lot of commits ahead

paulbdavis:
Noted, closing this and opening another PR on the develop branch

PR #192, opened 2019-12-14
allow config to disable go key while playing

paulbdavis:
Adds a config option to allow disabling the Go hotkey in list layout mode while there are actively playing cues. 

The Go button itself will still work, as will any programmed triggers, etc. It only disables the go key combo as to prevent accidentally triggering the next cue.

For some reason I cannot get the setting from the defaults to actually apply when a list layout is loaded (having to manually apply the settings via the layout menu), but this seems to be the case for all of the options at the moment

### DIFF
--- a/lisp/plugins/list_layout/layout.py
+++ b/lisp/plugins/list_layout/layout.py
@@ -37,7 +37,6 @@ from lisp.plugins.list_layout.view import ListLayoutView
 from lisp.ui.ui_utils import translate
 from lisp.ui.widgets.hotkeyedit import keyEventKeySequence
 
-
 class ListLayout(CueLayout):
     NAME = QT_TRANSLATE_NOOP("LayoutName", "List Layout")
     DESCRIPTION = QT_TRANSLATE_NOOP(
@@ -54,6 +53,7 @@ class ListLayout(CueLayout):
     ]
     Config = DummyConfiguration()
 
+    go_key_disabled_while_playing = ProxyProperty()
     auto_continue = ProxyProperty()
     dbmeters_visible = ProxyProperty()
     seek_sliders_visible = ProxyProperty()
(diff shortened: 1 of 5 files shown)


FrancescoCeruti:
> For some reason I cannot get the setting from the defaults to actually apply when a list layout is loaded (having to manually apply the settings via the layout menu), but this seems to be the case for all of the options at the moment

It's because the options you see in the layout menu (except "Show resize handles") are saved into the session file, this allow to have different saves with different options without changing them every time you switch file. The corresponding "defaults" only applied when you start a new session.

```python
# This expose the "option" to be saved into the session-file
go_key_disabled_while_playing = ProxyProperty()
    
# This two are defining how "go_key_disabled_while_playing" get/set it's value
@go_key_disabled_while_playing.set
def _set_go_key_disabled_while_playing(self, enable):
    self.go_key_disabled_while_playing_action.setChecked(enable)

@go_key_disabled_while_playing.get
def _get_go_key_disabled_while_playing(self):
    return self.go_key_disabled_while_playing_action.isChecked()
```

lisp/plugins/list_layout/layout.py
```diff
@@ -257,7 +266,10 @@ def _key_pressed(self, event):
             sequence = keyEventKeySequence(event)
             if sequence in self._go_key_sequence:
                 event.accept()
-                self.__go_slot()
+                if self._running_model.is_playing() is not True:
+                    self.__go_slot()
+                if self._get_go_key_disabled_while_playing() is not True:
+                    self.__go_slot()
```
FrancescoCeruti: I think it should be something like this:
```python
if not self.go_key_disabled_while_playing or not self._running_model.is_playing():
    self.__go_slot()
```

paulbdavis: I was debating this, thought this looked cleaner in terms of line length, but I forgot about returning explicitly so the next case is not checked. 

I'll get this tweaked and test out my changes given the notes you've made above

Issue #137, opened 2018-11-23
Drag & Drop to add new media cues


mitscherdinger:
Hi, there!

Using mate under Debian Testing, I just can't choose some audio files from the file manager and drop them into LSP. Could be a mate issue, but I thought I just keep you informed (in case it is not a mate issue).

ghassenghouibi:
Hello @FrancescoCeruti ,
Is this issue have been resolved ?
I tried to drag and drop audio file on the develop branch  but it doesn't work.I checked the files and i saw there is something to do (line commented #todo). So  i would like to work on this part but i don't know if it's already done.
Thank you

FrancescoCeruti:
Hi, it's not yet implemented, and I'm not working on it :smile:

PR #193, opened 2019-12-15
Drag'n drop files from file manager (issue #137)

frederic-peters:
Hi,
I was looking to use linux-show-player in our local radio and missing the ability to drop files from the file manager was a noted sore point, hence this pull request. It's made of two commits, one for the list view (this was our primary demand) and one for the card layout.
I only tested it with Nautilus but I don't think it has anything specific and it should work as is with other file managers. (also I never used PyQt before)

### DIFF
--- a/lisp/plugins/cart_layout/page_widget.py
+++ b/lisp/plugins/cart_layout/page_widget.py
@@ -21,6 +21,7 @@ from PyQt5.QtCore import pyqtSignal, Qt, QPoint
 from PyQt5.QtWidgets import QWidget, QGridLayout, QSizePolicy
 from sortedcontainers import SortedDict
 
+from lisp.backend import get_backend
 from lisp.core.util import typename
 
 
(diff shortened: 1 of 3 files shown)


frederic-peters:
This needs to be redone over the develop branch.

frederic-peters:
It has now been rebased over the develop branch; I also factorized some common code to gst_backend.

FrancescoCeruti:
Nice :-)
At a first look the code is good to me, I'm going to test it, probably tomorrow.

Thanks!

lisp/plugins/cart_layout/page_widget.py
```diff
@@ -99,10 +101,22 @@ def contextMenuEvent(self, event):
...
+                event.ignore()
+        else:
+            event.ignore()
 
     def dropEvent(self, event):
+        if event.mimeData().hasUrls():
+            # If files are being dropped, add them as cues
+            get_backend().cue_audio_urls(event.mimeData().urls())
+            return
+
```
FrancescoCeruti: This condition change significantly the way the function behave, it would be better to use an `else` statement instead of the `return`, it will make the flow more clear.

lisp/plugins/gst_backend/gst_backend.py
```diff
@@ -120,7 +121,21 @@ def _add_uri_audio_cue(self):
         if files:
             GstBackend.Config["mediaLookupDir"] = os.path.dirname(files[0])
             GstBackend.Config.write()
-
+        self._cue_audio_files(files)
+
+    def cue_audio_urls(self, urls):
```
FrancescoCeruti: I would name this `add_cue_from_urls`, it makes (at least to me) the use of the method more clear.

lisp/plugins/gst_backend/gst_backend.py
```diff
@@ -120,7 +121,21 @@ def _add_uri_audio_cue(self):
...
+        for url in urls:
+            filename = url.path()
+            basename, ext = os.path.splitext(url.fileName())
+            if ext.strip('.') not in extensions['audio'] + extensions['video']:
+                # Skip unsupported media types
+                continue
+            files.append(filename)
+        self._cue_audio_files(files)
+
+    def _cue_audio_files(self, files):
```
FrancescoCeruti: I would name this `add_cue_from_files`, without `_`, it could come useful as "public"

lisp/plugins/gst_backend/gst_backend.py
```diff
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Linux Show Player.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
```
FrancescoCeruti: looks like this is unused, `os.path` is already imported.

frederic-peters:
Thanks for the review, I made the suggested changes to my branch; I also did the return → else change onto lisp/plugins/list_layout/list_view.py.

PR #194, opened 2019-12-15
Add ctrl-q as shortcut to quit application

frederic-peters:
Hi,
I am used to Ctrl-Q for quitting an application and that shortcut was not implemented, so here is the obvious/trivial patch.
Thanks,


FrancescoCeruti:
Hey, looks fine to me, but you should work with the development branch.
I'm not going to merge new features in the current master, until the development branch is merged, and it's a lot of commits ahead.

Also you should use `QKeySequence.Quit` instead of a fixed value.

frederic-peters:
Thanks for the review; I missed the develop branch; I rebased mine and changed it to use QKeySequence.Quit.

PR #196, opened 2020-01-23
Implement a form of "plugin status"

s0600204:
This PR adds a way of acquiring the status or state of a plugin.

It is now possible to query a plugin directly to see if it is loaded or disabled; and to request a suitable status icon and tooltip for use in GUI ([preview](https://user-images.githubusercontent.com/2053873/73013176-ba584500-3e0f-11ea-988c-d9cb32f81b1b.jpg)).

There are also a few more status icon varieties.

----

Thoughts:

* Would it be worth adding methods to the `Plugin` class that can be used to check if it is in Error or Warning?
    - If Plugin-A could optionally use Plugin-B, then Plugin-A can simply check if Plugin-B is loaded or not. The fact that Plugin-B is in error is (for now) irrelevant.
    - Whilst the only current error is that dependencies are not satisfied (and this also prevents the plugin from being loaded), it is feasible that an error could occur after load. (Thus a plugin is both Loaded and Errored, in which case Plugin-A should need to detect that so it doesn't use Plugin-B.)
    - Perhaps an `is_usable` method? (`True` when Loaded and not Errored.)

* Should the Plugin class set all status flags internally (as with this PR the `DependenciesNotSatisfied` and `OptionalDependenciesNotSatified` flags are set outside the class)?

* Would the `is_disabled` method be better as an `is_enabled` method?

### DIFF
--- a/lisp/core/plugin.py
+++ b/lisp/core/plugin.py
@@ -17,6 +17,7 @@
 
 from lisp.core.configuration import DummyConfiguration
 from lisp.plugins import PluginState
+from lisp.ui.ui_utils import translate
 
 
 # TODO: add possible additional metadata (Icon, Version, ...)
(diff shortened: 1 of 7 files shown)


FrancescoCeruti:
Looks interesting

About your thoughts:

> Would it be worth adding methods to the Plugin class that can be used to check if it is in Error or Warning?

Not really, one could simply check the `Plugin.State` value, if needed, I think is better this way, since anyway is already there :)

> Should the Plugin class set all status flags internally (as with this PR the DependenciesNotSatisfied and OptionalDependenciesNotSatified flags are set outside the class)?

Not really sure what you mean, I suppose you suggest if it would be better to have methods to change the state? In that case is the same as the previous point

> Would the is_disabled method be better as an is_enabled method?

In my opinion yes, most of the time you do would do `if plugin.is_enabled():`, is a bit more concise and clear then `if not plugin.is_disabled():`.

PR #198, opened 2020-02-10
[0.5.x] Detect qt5ct and workaround it

s0600204:
(Written in response to the recent various "icons not being drawn" comments.)

`qt5ct` creates (or requires the user to create) the environment argument `QT_QPA_PLATFORMTHEME` with the value `qt5ct`.

The code supplied below detects the existence of the environment argument, checks its value, and if relevant adds a (blank) argument to be passed to the invocation of `QApplication`. This appears to causes Qt to ignore the environment value and fallback on the default platform theme.

----

A fix doesn't seem to be needed for the `0.6`/`develop` branch.

### DIFF
--- a/lisp/main.py
+++ b/lisp/main.py
@@ -21,9 +21,9 @@ import argparse
 import logging
 import sys
 from itertools import chain
-from os import path
+from os import environ, path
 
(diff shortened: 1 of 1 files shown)


FrancescoCeruti:
Nice, I'll look into merge it.
Did you consider using `os.getev()`?

s0600204:
Knew I was overlooking something...

PR #201, opened 2020-04-10
Support loading plugins from a User's home directory

s0600204:
This set of changes is an attempt to permit the above.

With this PR, it is possible to place a plugin in the folder `$XDG_DATA_HOME/LinuxShowPlayer/$LiSP_Version/plugins/` (e.g. `~/.local/share/LinuxShowPlayer/0.6/plugins/`) and LiSP will load it on next program restart.

As an example, I've created a plugin over at https://github.com/s0600204/protocol_monitor. To test, you can clone the repo into `~/.local/share/LinuxShowPlayer/0.6/plugins/protocol_monitor` (such that that is the root folder of the repo) and (re)start LiSP. (Or alternatively, move one of the plugins included with LiSP out.)

It is also possible with this PR to have a plugin zipped up. Unfortunately, downloading a zip-file from GitHub and using it directly is not possible (GitHub doesn't create the subfolder within the zip-file with the correct name), however creating a zip-file manually with the layout:
```
  example.zip
     '- example_plugin/
         |- __init__.py
         |- default.json
         '- example_plugin.py
```
should work.

There is a caveat with using a zip-file: whilst python code is included automatically, other files have to be extracted programmatically before being usable. The `default.json` file is example of this, and this pull-request includes code that creates a temporary folder and extracts it automatically on plugin load (the temporary folder is removed again when LiSP closes).

### DIFF
--- a/lisp/plugins/__init__.py
+++ b/lisp/plugins/__init__.py
@@ -17,7 +17,8 @@
 
 import inspect
 import logging
-from os import path
+from os import makedirs, path
+import sys
 
 from lisp import app_dirs
 from lisp.core.configuration import JSONFileConfiguration
@@ -28,6 +29,11 @@ PLUGINS = {}
 LOADED = {}
 
 FALLBACK_CONFIG_PATH = path.join(path.dirname(__file__), "default.json")
+USER_PLUGIN_PATH = path.join(app_dirs.user_data_dir, "plugins")
+
+# Make sure the path exists, and insert it into the list of paths python uses to find modules
+makedirs(USER_PLUGIN_PATH, exist_ok=True)
+sys.path.insert(1, USER_PLUGIN_PATH)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


FrancescoCeruti:
So, I've looked at it, the user directory part seems fine to me :+1:

I'm not sure about the zip loading, not the code itself, but as a feature and how useful it can be, considering that is also have a few limitations.

I think that my original idea was to provide the ability to install third-party plugins via zip file using some kind of functionality (with a UI) that would extract the archive and hide a few steps from the user.

This second approach would have the following advantages:
* the user doesn't need to know were the plugins are located
* no special concerns about non-python files (resources)
* it's possible to check for the zip structure during the installation
* later we can provide other checks (e.g. version compatibility)
* the plugin discovery can be kept slighter simpler

Since currently I don't think there will be many users for this feature we can simply add support for loading the plugins from the user directory, and at a later point we could provide all the functionality mentioned above.

This is my opinion obviously, maybe I'm missing something, if you have some counter arguments let me know :)

s0600204:
Fair enough. In which case, only the first commit is actually needed.

I'll rebase and push that so it's easier to review.

PR #202, opened 2020-04-30
Draw scale markings on level meters

s0600204:
The markings shown are those determined (by the code) to fit the space available without overlapping.

Examples:
Meter Marks - Cart Meter Marks - List Meter Marks - RF

### DIFF
--- a/lisp/ui/widgets/dbmeter.py
+++ b/lisp/ui/widgets/dbmeter.py
@@ -17,8 +17,8 @@
 
 from typing import Callable
 
-from PyQt5.QtCore import QRect, QRectF
-from PyQt5.QtGui import QLinearGradient, QColor, QPainter, QPixmap
+from PyQt5.QtCore import QPointF, QRect, QRectF, Qt
+from PyQt5.QtGui import QLinearGradient, QColor, QPainter, QPixmap, QFontMetrics
 from PyQt5.QtWidgets import QWidget
 
 from lisp.backend.audio_utils import iec_scale
@@ -35,6 +35,7 @@ class DBMeter(QWidget):
         clipping: int = 0,
         smoothing: float = 0.66,
         scale: Callable = iec_scale,
+        unit: str = 'dBFS',
         **kwargs,
     ):
         super().__init__(parent, **kwargs)
@@ -43,13 +44,22 @@ class DBMeter(QWidget):
         self.dbClipping = clipping
         self.valueSmoothing = smoothing
         self.scale = scale
+        self.unit = unit
 
         self.backgroundColor = QColor(32, 32, 32)
         self.borderColor = QColor(80, 80, 80)
         self.clippingColor = QColor(220, 50, 50)
 
         self._currentSmoothing = self.valueSmoothing
+        self._markings = []
         self._pixmap = QPixmap()
+
+        font = self.font()
+        font.setPointSize(font.pointSize() - 2)
+        self.setFont(font)
+        font.setPointSize(font.pointSize() - 2)
+        self.unit_font = font
+        self.scale_width = 0
 
         self.reset()
 
@@ -77,6 +87,35 @@ class DBMeter(QWidget):
 
         self.update()
 
+    def updateMarkings(self):
+        self._markings = []
+
+        height = self.height()
+        font_height = QFontMetrics(self.font()).height()
+        steps = [1, 5, 10, 25, 50, 100, 250]
+        curr_level = self.dBMax
+        curr_y = 0
+
+        while curr_y < height - font_height:
+            prev_level = curr_level
+            prev_y = curr_y + font_height
+
+            for step in steps:
+                curr_level = prev_level - step
+                curr_y = height - self.scale(curr_level) * (height - 2)
+                if curr_y > prev_y:
+                    break
+
+            self._markings.append([curr_y, curr_level])
+
+        self._markings.pop()
+
+        self.scale_width = max(
+            QFontMetrics(self.font()).boundingRect(str(-abs(self.dBMax - self.dBMin))).width(),
+            QFontMetrics(self.unit_font).boundingRect(self.unit).width()
+        )
+
+
     def updatePixmap(self):
         """Prepare the colored rect to be used during paintEvent(s)"""
         w = self.width()
@@ -95,6 +134,7 @@ class DBMeter(QWidget):
 
     def resizeEvent(self, event):
         self.updatePixmap()
+        self.updateMarkings()
 
     def paintEvent(self, e):
         height = self.height()
@@ -105,7 +145,7 @@ class DBMeter(QWidget):
         painter.setBrush(self.backgroundColor)
 
         # Calculate the meter size (per single channel)
-        meterWidth = width / len(self.peaks)
+        meterWidth = (width - self.scale_width) / len(self.peaks)
         meterRect = QRect(0, 0, meterWidth - 2, height - 1)
 
         # Draw each channel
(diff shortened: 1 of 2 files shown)


FrancescoCeruti:
Nice! I think I've a few improvements to suggest, I need to figure out a couple of things, then I'll make a revision to highlight them :)

lisp/ui/widgets/dbmeter.py
```diff
@@ -43,14 +44,23 @@ def __init__(
...
         self.backgroundColor = QColor(32, 32, 32)
         self.borderColor = QColor(80, 80, 80)
         self.clippingColor = QColor(220, 50, 50)
 
         self._currentSmoothing = self.valueSmoothing
+        self._markings = []
         self._pixmap = QPixmap()
 
+        font = self.font()
+        font.setPointSize(font.pointSize() - 2)
```
FrancescoCeruti, line 58: I think we should use a fixed/monospaced font here, it should improve readability, and make the character size more consistent. In this case I've also noticed that's better to make it a bit smaller

```suggestion
        font = QFontDatabase.systemFont(QFontDatabase.FixedFont)
        font.setPointSize(font.pointSize() - 3)
```

lisp/ui/widgets/dbmeter.py
```diff
@@ -77,6 +87,35 @@ def plot(self, peaks, _, decayPeak):
 
         self.update()
 
+    def updateMarkings(self):
+        self._markings = []
+
+        height = self.height()
+        font_height = QFontMetrics(self.font()).height()
```
FrancescoCeruti, line 94: To optimize the space we can "remove" some empty space on top of the numbers, using `QFontMetrics.descent()` should provide the size of this value

```suggestion
        fm = QFontMetrics(self.font())
        font_height = fm.height() - fm.descent()
```

lisp/ui/widgets/dbmeter.py
```diff
@@ -77,6 +87,35 @@ def plot(self, peaks, _, decayPeak):
...
+        steps = [1, 5, 10, 25, 50, 100, 250]
```
FrancescoCeruti, line 95:
```suggestion
        steps = [2, 4, 8, 16, 32, 64, 128]
```
I found that this scale is a bit more pleasing, but le me know what do you think :)

lisp/ui/widgets/dbmeter.py
```diff
@@ -77,6 +87,35 @@ def plot(self, peaks, _, decayPeak):
...
+        curr_y = 0
+
+        while curr_y < height - font_height:
+            prev_level = curr_level
+            prev_y = curr_y + font_height
+
+            for step in steps:
+                curr_level = prev_level - step
+                curr_y = height - self.scale(curr_level) * (height - 2)
+                if curr_y > prev_y:
```
FrancescoCeruti, line 106: We should take in account that the text will be vertically centered
```suggestion
                if curr_y - ceil(font_height / 2) > prev_y:
```

lisp/ui/widgets/dbmeter.py
```diff
@@ -77,6 +87,35 @@ def plot(self, peaks, _, decayPeak):
...
+                    break
+
+            self._markings.append([curr_y, curr_level])
+
+        self._markings.pop()
+
+        self.scale_width = max(
+            QFontMetrics(self.font()).boundingRect(str(-abs(self.dBMax - self.dBMin))).width(),
```
FrancescoCeruti, line 114: I suppose we could assume that self.dBMin will be the "longest" value?
```suggestion
            fm.boundingRect(str(self.dBMin)).width(),
```

lisp/ui/widgets/dbmeter.py
```diff
@@ -143,7 +183,26 @@ def paintEvent(self, e):
             )
             painter.drawPixmap(decayRect, self._pixmap, decayRect)
 
+            # Draw markings
+            x_start = meterRect.x() + meterRect.width() / 2
+            x_end = meterRect.x() + meterRect.width()
+            for mark in self._markings:
+                painter.drawLine(QPointF(x_start, mark[0]), QPointF(x_end, mark[0]))
+
```
FrancescoCeruti, line 191: This should be moved before drawing the "decay indicator", or sometime you might see only the left half of the indicator

lisp/ui/widgets/dbmeter.py
```diff
@@ -143,7 +183,26 @@ def paintEvent(self, e):
...
             # Move to the next meter
             meterRect.translate(meterWidth, 0)
 
+        # Write the scale marking text
+        text_height = QFontMetrics(self.font()).height()
+        text_offset = text_height / 2
+        painter.setPen(self.palette().windowText().color())
+        painter.drawText(0, 0, width, text_height, Qt.AlignRight, str(self.dBMax))
```
FrancescoCeruti, line 199: We can account for the empty space on top of the character, using (as before) `QFontMetrics.descent()`
```suggestion
        painter.drawText(0, -QFontMetrics(self.font()).descent(), width, text_height, Qt.AlignRight, str(self.dBMax))
```

lisp/ui/widgets/dbmeter.py
s0600204: Intentional.

lisp/ui/widgets/dbmeter.py
s0600204: This exact change would have the sole result of drawing the top half of the value of `self.dBMax` off the top of the widget, clipping it.

lisp/ui/widgets/dbmeter.py
s0600204: Not necessarily - the third image in the PR description shows an example where that's not the case: `self.dBMin` = 0, `self.dBMax` = 40

lisp/ui/widgets/dbmeter.py
s0600204: Done. Differently, but done.

lisp/ui/widgets/dbmeter.py
s0600204: The scale given is designed to give values that one might expect to find on, say, the meter-bridge of a sound desk.

I'll grant you, that some revision was needed: a gap of 100 dB was implausible.

lisp/ui/widgets/dbmeter.py
s0600204: According to the spec:
> `QFontMetrics::height()` [...] is always equal to `ascent()`+`descent()`.

So, we can use `fm.ascent()`, and save ourselves a calculation.

lisp/ui/widgets/dbmeter.py
s0600204: Done

lisp/ui/widgets/dbmeter.py
FrancescoCeruti: eheh, sorry, I've looked at it forgot :facepalm:

lisp/ui/widgets/dbmeter.py
FrancescoCeruti: :thinking: I get a different result, for some reason I've a larger gap at the top, probably depends a lot from the used font, better not to use it

lisp/ui/widgets/dbmeter.py
FrancescoCeruti: make sense, for some reason I find it a bit annoying, but probably in normal conditions is not really noticeable :+1:

FrancescoCeruti:
> Hmm. I've given it a go, and it now shouldn't repaint the labels when plot() is called. It'll still repaint them more frequently than strictly necessary perhaps, but meh.

okay, I'll look into it myself :P

PR #206, opened 2020-11-01
Implement ability for set a preferred default ALSA device

s0600204:
The chosen ALSA device will be used by new (Gst)Media Cues, instead of using the system's default device.

Previously if a user wished to use - for example - a USB soundcard, then the user would have to manually set that in every cue. Now the user may set it once, globally, and all (Gst)Media cues created afterwards will use that device.

----

The new setting may be found/configured under File > Preferences; Plugins > GStreamer > ALSA Default Device

### DIFF
--- a/lisp/plugins/gst_backend/settings/alsa_sink.py
+++ b/lisp/plugins/gst_backend/settings/alsa_sink.py
@@ -1,6 +1,6 @@
 # This file is part of Linux Show Player
 #
-# Copyright 2018 Francesco Ceruti <ceppofrancy@gmail.com>
+# Copyright 2020 Francesco Ceruti <ceppofrancy@gmail.com>
 #
 # Linux Show Player is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


lisp/plugins/gst_backend/settings/alsa_sink.py
```diff
@@ -76,10 +73,12 @@ def enableCheck(self, enabled):
         self.setGroupEnabled(self.deviceGroup, enabled)
 
     def loadSettings(self, settings):
-        device = settings.get("device", "default")
+        device = settings.get("device", "")
+        if not device:
+            device = get_plugin('GstBackend').Config.get('alsa_device', AlsaSink.FALLBACK_DEFAULT_DEVICE)
```
FrancescoCeruti, line 78: Same as before, we can use `GstBackend` directly, and probably just pass the config value as default?

```suggestion
        device = settings.get("device", GstBackend.Config.get('alsa_device', AlsaSink.FALLBACK_DEFAULT_DEVICE))
```

Issue #69, opened 2017-03-18
[Enhancement] Edit columns (add/remove)


GodsVictory:
I'm currently running a show using this software, so thank you! The tracks we use have the numbers in the filenames, so having a second track number is redundant, and confusing when they don't line up the same. It would be nice to be able to hide the track number column.

FrancescoCeruti:
Hi, I'm already planning to add new options for the columns, with that even the ability to show/hide (move?) columns. Unfortunately I think it will take some time before we can see this feature.

PR #210, opened 2021-06-28
Edit columns 

Rob7000:
Proposed change for #69 : 
- user can drag columns to change their order. 
- context menu on header to allow each column to be hidden
- when columns are hidden, the context menu then provides list of headers to be shown.

As my first contribution, please let me know how it can be improved.


FrancescoCeruti:
This is now implemented in 592ef069

PR #213, opened 2021-11-06
Restore background to `RunningCueWidget` instances

s0600204:
Recent commit 6fa0bfda0453d tweaked the dark-theme palette. This resulted in the List Layout's `RunningCueWidget`s to change appearance as depicted below (left: before; right: after):

Before and After

(Seems to be a side-effect of removing the `background-color` rule from the `QWidget` ruleset within the stylesheet.)

I'm guessing that this change to the background of the `RunningCueWidget` wasn't intended, so this PR restores it:

Result of PR Changes

You might notice that the new background is `palette(mid)` instead of `palette(window)` (the latter would be closer to the appearance pre- 6fa0bfda0453d). Feel free to change that if you prefer.

### DIFF
--- a/lisp/ui/themes/dark/theme.qss
+++ b/lisp/ui/themes/dark/theme.qss
@@ -717,6 +717,10 @@ CueListView:focus {
     border: 1px solid palette(light);
 }
 
+RunningCueWidget>QWidget {
+    background-color: palette(mid);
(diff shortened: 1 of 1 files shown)


FrancescoCeruti:
Thank you!

PR #229, opened 2022-02-20
Remove the OSC cue's 'Pause' action

s0600204:
As it is no longer possible (since f937bde800a7) for `Fader`s to be paused, it is also no longer possible to pause running (e.g. fading) OSC cues.


FrancescoCeruti:
I knew I missed something :upside_down_face: 

Thanks!

Issue #241, opened 2022-05-21
Show filename/title in the window titlebar


rdealmeida12:
I do sound effects for a local theater group here in Florida.  Linux show player works great and several other groups have asked about it.
Can you add a title line to Cart Layout ?

Thanks
Bob

FrancescoCeruti:
Hi! do you mean custom titles in the tabs like this?
 
![image](https://user-images.githubusercontent.com/5596673/169699151-352ce80d-9e83-495c-86ea-bc147e404719.png)

It will be included in the 0.6 release

rdealmeida12:
Here is a screenshot of the effects for "The Aldrich Family" OTR radio play we just did. I would be great if title of the .lsp were displayed somewhere on to
![linux show player](https://user-images.githubusercontent.com/8337231/169709238-b79a8d37-aa12-4d66-bfbf-40ec5a2ae392.png)
p

FrancescoCeruti:
Oh, okay, in the window title, that should also already be in the next version.

I still need to polish a few things. But you can already try the 0.6 "experimental" version https://www.linux-show-player.org/download.html keep in mind that the save format is incompatible with the previous versions.

rdealmeida12:
Thank You. I'll check out the new version and let you know

Thanks
Bob

PR #243, opened 2022-05-25
Update window title on file creation and load

s0600204:
Previously the window title was only updated when an open file was saved, a cue was edited, or an undo/redo occurred.

This had interesting side effects, such as when one showfile was saved and the user then opens a second: LiSP's titlebar still showed the name of the saved file and not that of the just-loaded file.

Related to #241


FrancescoCeruti:
Thanks :smile:

FrancescoCeruti:
Hey I've noticed the issue was not solved when the session was loaded from the layout selection dialog, on startup, or via "new session". It's fixed here, if you're interested :wink:  7eb4ed5af9376ca4051babaf39e5d25c41b229e8

Issue #235, opened 2022-03-02
MIDI note_off issue


moonshaadow:
**Describe the bug**
New midi features are really nice.
But a strange behavior is always here.
When I send a note_off message, LiSP read it as a note_on.

Then if I try to manually set a note_off message, this is not working as expected.

**To Reproduce**

1. Go to Edit Cue > Cue Control > Midi Controls
2. Click on capture
3. send a midi note_on then a midi_note_off
4. See that  both two messages are set to "NOTE_ON"

OR:
1. Add manually a note_on midi control and set it to START (not DEFAULT)
2. Add manually a note_off midi control with same datas and set it to STOP (not DEFAULT)
3. test the behavior of the midi note and see that note_on action is ok but note_off action does nothing

**Expected behavior**
NOTE_OFF should be correctly recognized and should trigger action .

**Screenshots**
![Capture d’écran du 2022-03-02 23-55-09](https://user-images.githubusercontent.com/29754932/156463620-b7aebce5-ea83-4ea2-b0de-681a3dabd1f0.png)

**Linux Show Player version:**
built from development branch

**OS Version**
Debian Testing

FrancescoCeruti:
I'm unable to reproduce the behavior. Can you tell me how did you send the midi messages?

MIDI is mostly handled via an external library (mido), maybe it's related to this https://github.com/mido/mido/issues/130 ?

moonshaadow:
I send the messages with a midi keyboard, I could try with others tools if you want. But I trust the output of Gmidimonitor, which is good for this task. I've tried with another monitor as well.

FrancescoCeruti:
I'm testing using https://vmpk.sourceforge.io/ and monitoring with Gmidimonitor or https://kmidimon.sourceforge.io/ I get the same readings that on the application.

The only exception is the one mentioned before in the link (note_on with 0 velocity might be interpreted differently).

I might try to build a VM to test on Debian :thinking:

Stav88:
I reproduce the bug using material (ProdipeMIDIlilo). So I try to test with VMPK but it doesn't produce the bug. It seems the it need material midi.

Tested OS : 
 - Xubuntu 20.04 x64 (both material & VMPK)
 - Manjaro Aarch64 (material only)

Tested application versions : 
 - 0.5.1 (Xubuntu)
 - master (Xubuntu)
 - 0.5.3 (Manjaro)

I can try to debug on master.

PR #256, opened 2022-08-29
#235 Midi message: process Note On with Velocity=0 as a Note Off

Stav88:
As Midi specification said :
> A device that recognizes MIDI Note On messages must be able to recognize both a real Note Off as well as a Note On with 0 velocity (as a Note Off). There are many devices that generate real Note Offs, and many other devices that use Note On with 0 velocity as a substitute.

I add debug messages to debug midi messages filtering "clock" message to avoid flood.


Stav88:
Sorry bad issue referenced

PR #257, opened 2022-08-29
#235 Midi message: process Note On with Velocity=0 as a Note Off

### LINKED ISSUE #235: MIDI note_off issue

Stav88:
fix #235 
As Midi specification said :
> A device that recognizes MIDI Note On messages must be able to recognize both a real Note Off as well as a Note On with 0 velocity (as a Note Off). There are many devices that generate real Note Offs, and many other devices that use Note On with 0 velocity as a substitute.

I add debug messages to debug midi messages filtering "clock" message to avoid flood.

### DIFF
--- a/lisp/modules/midi/midi_input.py
+++ b/lisp/modules/midi/midi_input.py
@@ -20,6 +20,8 @@
 from lisp.core.signal import Signal
 from lisp.modules.midi.midi_common import MIDICommon
 from lisp.modules.midi.midi_utils import mido_backend, mido_port_name
+import logging
+from mido.messages import Message
 
 
 class MIDIInput(MIDICommon):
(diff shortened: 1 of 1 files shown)


FrancescoCeruti (changes requested):
This might be a breaking change for someone, so it should target the "develop" branch, the file should be `lisp/plugins/midi/midi_io.py`.

FrancescoCeruti:
I've merged the fix into the dev version, thanks!

PR #280, opened 2023-08-07
Implement a Loop-Release CueAction

s0600204:
This PR adds a new "Loop Release" action to Media Cues.

When triggered on a cue that is looping, it causes the current loop to finish before stopping the cue; in comparison to the "Stop" action which stops the cue straight away (with or without a fade).

I personally use it on shows for things such as phone ringing sound effects (which sound odd if stopped abruptly); whilst a recent user on the gitter chat mentioned that this sort of functionality might be useful for "vamping": where a short stretch of music is looped for a period of time that varies between performances (for instance: underscoring the dialogue between verses in a song).

### DIFF
--- a/lisp/cues/cue.py
+++ b/lisp/cues/cue.py
@@ -502,6 +502,10 @@ class Cue(HasProperties):
         :type fade: bool
         """
 
+    def loop_release(self):
+        """Release any remaining cue loops."""
+        pass
+
     def fadein(self, duration, fade_type):
         """Fade-in the cue.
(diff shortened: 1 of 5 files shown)


FrancescoCeruti:
I've added a few bits:

1. check on the media status, to avoid releasing stopped cues, which can be an unexpected behaviour
2. add the LoopRelease action to the cue "stop" options in the UI
3. add the method signature to the base Cue class, with a short docstring

PR #326, opened 2024-08-19
Correctly save index column visibility

s0600204:
At the moment, the "Show Index Column" option (in the "Layout" menu) is saved inverted to what it's set.

This means that if the Index Column is shown and the user then "saves" the showfile, the Column is saved as hidden, causing the column to not be shown when the showfile is next loaded. The reverse is also true.

This PR corrects this.


FrancescoCeruti:
Thanks! :+1:

PR #331, opened 2024-10-12
Port to PyQt6

FrancescoCeruti:


a17r:
### Qt5 is now declared deprecated in Gentoo

At this point in time, Linux distributions are already tracking which packages are blocking a future Qt5 cleanup. Real Qt Company OSS support stopped back in 2020, when public availability of LTS releases started to be delayed for 1 year, stable branch repository access got closed down (with 5.15 branch repository receiving any new LTS tag one year after commercial release) and only Qt6-relevant bugs are being fixed (then backported) since then. The final release happened on 2025-05-19, which means a) we will not get it until 2026-06-19, b) any bugs discovered since last May will not be fixed and c) no bugs since 2020-12-08 that haven't also been affecting Qt6 were fixed at all.

*At least* Gentoo, Alpine and Ubuntu are entertaining Qt6 porting trackers by now, for Gentoo, Qt5 revdep cleanups are in full swing already.
https://bugs.gentoo.org/show_bug.cgi?id=qt5-removal
https://gitlab.alpinelinux.org/alpine/aports/-/issues/17114
https://ubuntu-archive-team.ubuntu.com/transitions/html/qt5-removal.html

PR #333, opened 2024-12-11
Basic Video support

essboyer:
# Basic Video Support

This PR provides a working POC (proof-of-concept) demonstrating native video playback capabilities inside Linux Show Player. In the final phase it will provide a basic implementation of video support which should be good enough for many shows (and which should ultimately address issue #3). 

In the POC Phase (Phase I) LSP uses the WaylandSink for full-sized video playback on the main display. The video can be set to play in `fullscreen`. In Phase II a choice of video sinks may be presented to suit the environment.

Code changes are entirely within the `gst_backend` plugin.

## What's working

- add a video cue (right-click context menu)
- add a video cue (drag & drop)
- video playback on **main display only**
- fullscreen playback on main display
- seek/pause/etc
- audio playback
- audio pipeline (db meters, etc)
- video duration

## TODO
- video playback on external monitor/projector
- more robust video sink selection
- hide mouse during playback
- fade video opacity
- adjust other video parameters (brightness, contrast, etc)
- image transform (keystone)
- video playback in a window (audition)
- video thumbnail

### DIFF
--- a/lisp/plugins/gst_backend/default.json
+++ b/lisp/plugins/gst_backend/default.json
@@ -1,5 +1,7 @@
 {
-  "_version_": "2",
+  "_version_": "3",
   "_enabled_": true,
+  "audio-pipeline": ["Volume", "Equalizer10", "DbMeter", "AutoSink"],
+  "video-pipeline": ["Volume", "DbMeter", "Equalizer10", "WaylandSink", "AutoSink"],
   "pipeline": ["Volume", "Equalizer10", "DbMeter", "AutoSink"]
 }
--- a/lisp/plugins/gst_backend/elements/wayland_sink.py
+++ b/lisp/plugins/gst_backend/elements/wayland_sink.py
@@ -1,0 +1,43 @@
+# This file is part of Linux Show Player
+#
+# Copyright 2024 Francesco Ceruti <ceppofrancy@gmail.com>
+#
+# Linux Show Player is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Linux Show Player is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Linux Show Player.  If not, see <http://www.gnu.org/licenses/>.
+
+from PyQt5.QtCore import QT_TRANSLATE_NOOP
+
+from lisp.backend.media_element import ElementType, MediaType
+from lisp.plugins.gst_backend.gi_repository import Gst
+from lisp.plugins.gst_backend.gst_element import GstMediaElement
+from lisp.plugins.gst_backend.gst_properties import GstProperty
+
+
+class WaylandSink(GstMediaElement):
+    ElementType = ElementType.Output
+    MediaType = MediaType.Video
+    Name = QT_TRANSLATE_NOOP("MediaElementName", "Wayland Video")
+
+    def __init__(self, pipeline):
+        super().__init__(pipeline)
+        
+        device = GstProperty("wayland_sink", "display", default="wayland-0")
+        fullscreen = GstProperty("wayland_sink", "fullscreen", default=True)
+
+        self.wayland_sink = Gst.ElementFactory.make("waylandsink", "wayland_sink")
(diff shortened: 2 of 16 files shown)


lisp/plugins/gst_backend/elements/wayland_sink.py
```diff
@@ -0,0 +1,43 @@
...
+from PyQt5.QtCore import QT_TRANSLATE_NOOP
+
+from lisp.backend.media_element import ElementType, MediaType
+from lisp.plugins.gst_backend.gi_repository import Gst
+from lisp.plugins.gst_backend.gst_element import GstMediaElement
+from lisp.plugins.gst_backend.gst_properties import GstProperty
+
+
+class WaylandSink(GstMediaElement):
+    ElementType = ElementType.Output
```
s0600204, line 27: 1. if a user opens a newly created Media (Video) Cue, then opens the "Edit Pipeline" dialog, it is possible to specify the "Wayland Video" element as the Output (instead of "System Out", "ALSA Out", ...), leading to (once the dialog closed) "Wayland Video" element appearing twice in the pipeline.

2. if a user removes the "Wayland Video" element from the pipeline of a Media (Video) Cue, then closes the "Edit Pipeline" dialog, upon reopening it, it is no longer possible to add the "Wayland Video" element back.

Both can be resolved by setting this to `ElementType.Plugin`.

Admittedly this is perhaps *conceptually* wrong, but it works!

lisp/plugins/gst_backend/default.json
```diff
@@ -1,5 +1,7 @@
 {
-  "_version_": "2",
+  "_version_": "3",
   "_enabled_": true,
+  "audio-pipeline": ["Volume", "Equalizer10", "DbMeter", "AutoSink"],
+  "video-pipeline": ["Volume", "DbMeter", "Equalizer10", "WaylandSink", "AutoSink"],
   "pipeline": ["Volume", "Equalizer10", "DbMeter", "AutoSink"]
```
s0600204, line 6: The `pipeline` key is still read and written by the application preferences dialog ("File" > "Preferences" then "Plugins" > "GstBackend"), where it is intended to be used to allow users to specifiy what the default pipeline for new Media (Audio) Cues should be.

However, as this key is no longer used by the `GstMediaCue` class to build the initial pipeline, this no longer works as intended.

s0600204 (commented):
Hello, and welcome to the project from a (fairly) long-time contributor!

Fiirstly, thank you for looking into this. Speaking anecdotally: over the past couple of years I've been involved in shows that have wanted to use video projection, _QLab_ has just not wanted to behave itself (stuttering and juddering, despite running on a computer that can play the files just fine via QuickTime), _ShowCueSystems_ hasn't worked reliably either, and all I've been able to do is shrug as my weapon of choice (LiSP) doesn't support outputting video at all. (And it seems like I cannot get my head round GStreamer at all, which is why I never got anywhere with this myself.) So seeing some progress in this area is much appreciated.

So, I don't have Wayland setup on the computer I use day-to-day, so this immediately doesn't work for me (the cue is marked as "playing" in the UI but doesn't progress and no audio is heard). But that's a me problem, rather than a you problem (And something I probably should look into, if only to know how to get it working on my "show" computer). My point is, I'd be happy to see other sinks implemented, if you get the time.

tobi08151405:
Hey, I wrote my own simple PoC for video support, mine is not so clean and only supports AutoVideoSink, but I found a solution for getting fading to work (see alpha.py plugin).

[https://github.com/tobi08151405/linux-show-player/tree/video-support](https://github.com/tobi08151405/linux-show-player/tree/video-support)

tobi08151405:
> Hello, and welcome to the project from a (fairly) long-time contributor!
> 
> Fiirstly, thank you for looking into this. Speaking anecdotally: over the past couple of years I've been involved in shows that have wanted to use video projection, _QLab_ has just not wanted to behave itself (stuttering and juddering, despite running on a computer that can play the files just fine via QuickTime), _ShowCueSystems_ hasn't worked reliably either, and all I've been able to do is shrug as my weapon of choice (LiSP) doesn't support outputting video at all. (And it seems like I cannot get my head round GStreamer at all, which is why I never got anywhere with this myself.) So seeing some progress in this area is much appreciated.
> 
> So, I don't have Wayland setup on the computer I use day-to-day, so this immediately doesn't work for me (the cue is marked as "playing" in the UI but doesn't progress and no audio is heard). But that's a me problem, rather than a you problem (And something I probably should look into, if only to know how to get it working on my "show" computer). My point is, I'd be happy to see other sinks implemented, if you get the time.

If it is still relevant for you, I now have a new merge-request open (#349 ) which adds the autovideosink to system out and should therefore not rely on wayland for any output, there also some other features implemented

Issue #335, opened 2025-01-01
Application crashes on save with bad data in session


essboyer:
**Describe the bug**
If you attempt to save a session containing data unserializable by JSON parsers, the application will crash.

**To Reproduce**
1. Create a list session
2. Add an audio cue
3. [bad plugin injects bad data]
4. Save session (ctrl+s)
5. Session file is truncated to 0
6. Application terminates

**Expected behavior**
 - Display an error to the user informing the session cannot be saved due to bad data
 - Advise user to check their cue pipelines
 - Do **not** overwrite the current saved session file
 - Application continues to run

**Screenshots**
N/A

**Linux Show Player version:**
- main

**OS Version**
 - Debian 12

**Additional context**
This is more of an improvement than a "bug", per se, since if the plugin developers have done their jobs, there should be no unparseable data in the in-memory session. However, the application should not crash due to the inability to save a sane session file. This fix allows the user to correct the issue that caused the problem (e.g removing a problematic plugin from a cue pipeline).

fnetX:
I would argue that the error message from https://github.com/FrancescoCeruti/linux-show-player/pull/336 is not "meaningful."

> Due to bad data, the session cannot be saved.
> Please check any recent changes to cue pipelines.

First of all, I am not sure what "cue pipelines" are exactly, and I fear that users will end in this situation:

- They did some modifications to a cue.
- They don't save immediately.
- They do further modifications to the showfile.
- They are unable to save due to an error message that doesn't help them.

First of all, I would suggest to reword the error message as a hotfix (explain what happened, then why, then recommend which actions to take). Usually, I'm a fan of putting the recommended action first so a user does not panic, but in this case, it is probably better to first explain the current situation, then elaborate on potential steps).

> Saving your session failed due to bad data. The saved file contains the last successful save and was not overwritten.
>
> This is likely to come from corrupt data streaming from a plugin. If you recently added any plugins to a cue, please try removing them and saving the file again.

Ideally, the file would still save properly but *without* the corrupt data, so that other unrelated settings are kept, only warning the user that certain settings are not properly saved, optionally asking for confirmation: "Save nevertheless. Some settings might not be saved correctly, but your file will be loadable." and "Do not overwrite last successful save. Your recent modifications are not saved".

FrancescoCeruti:
Thank @fnetX, I've improved the message (1aa60435ee08513da5c74bb7ac8fa79b627a185c).

[...]

PR #336, opened 2025-01-01
Improvement: Prevent application crash on Save Session

essboyer:
**Abstract**

This PR is meant to address issue #335 - Application crashes on save with bad session data.

**Description**

This PR changes the logic of `__save_to_file()` in `application.py` by first checking to ensure the JSON serialization of session data was successful before opening the file, writing data, changing application title and saving session.

In the event there is data present in the in-memory session which is **not** serializable into JSON (for example, if a plugin `GstProperty` returns a GObject rather than a simple type) a meaningful error message will be displayed to the user.

lsp-bad-session-data-error-message

### DIFF
--- a/lisp/application.py
+++ b/lisp/application.py
@@ -194,25 +194,36 @@ class Application(metaclass=Singleton):
             "cues": session_cues,
         }
         
-        # Verify session data before writing to file
-        session_json = None
         if self.conf.get("session.minSave", False):
             dump_options = {"separators": (",", ":")}
         else:
             dump_options = {"sort_keys": True, "indent": 4}
+
+        # Serialize the session in a JSON string, and overwrite the active session file
         try:
             session_json = json.dumps(session_dict, **dump_options)
         except TypeError:
             logger.exception(
-            translate(
-                "ApplicationError",
-                '''
-                Error while writing the session file "{}"\n
-                Due to bad data, the session cannot be saved.
-                Please check any recent changes to cue pipelines.
-                '''
-            ).format(session_file)
-        )
+                translate(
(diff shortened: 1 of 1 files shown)


lisp/application.py
```diff
@@ -193,23 +193,39 @@ def __save_to_file(self, session_file):
...
+            logger.exception(
+            translate(
+                "ApplicationError",
+                '''
+                Error while writing the session file "{}"\n
+                Due to bad data, the session cannot be saved.
+                Please check any recent changes to cue pipelines.
+                '''
+            ).format(session_file)
+        )
```
FrancescoCeruti, line 215: We can avoid creating the empty `session_json` using `try...else`

```suggestion
        
        if self.conf.get("session.minSave", False):
            dump_options = {"separators": (",", ":")}
        else:
            dump_options = {"sort_keys": True, "indent": 4}

        # Serialize the session in a JSON string, and overwrite the active session file
        try:
            session_json = json.dumps(session_dict, **dump_options)
        except TypeError:
            logger.exception(
                translate(
                    "ApplicationError",
                    """
                    Error while writing the session file "{}"\n
                    Due to bad data, the session cannot be saved.
                    Please check any recent changes to cue pipelines.
                    """,
                ).format(session_file)
            )
        else:
            with open(session_file, mode="w", encoding="utf-8") as file:
                file.write(session_json)

            # Save last session path
            self.conf.set("session.lastPath", dirname(session_file))
            self.conf.write()

            # Set the session as saved
            self.commands_stack.set_saved()
            self.window.updateWindowTitle()
```

PR #338, opened 2025-01-03
Replace (py)liblo(3)

s0600204:
This first commit of this PR removes `pyliblo`/`pyliblo3` (and its `liblo` dependency) and replaces it with `python-osc` (https://github.com/attwad/python-osc).

This has a number of advantages, namely:

1. `python-osc` is pure-python, thus removing a non-python dependency (and making cross-platform implementation of LiSP easier)
2. It's easier to obtain (than `pyliblo`) (at least from `pip`, if not currently from [system repositories](https://repology.org/project/python:python-osc/versions))
3. It's maintained (if not actively developed) (unlike `pyliblo`), and the versions are tagged (unlike `pyliblo3`)

The second commit rewrites the OSC plugin's application-level configuration page to (hopefully) make it easier to understand for users. Not strictly pertinent to replacing the OSC library, but as I was working in the area...

What isn't included in this PR:

* A rebuilt lockfile
* An updated `flatpak` specification


FrancescoCeruti:
@s0600204 ther's a little conflict, can you rebase?

Issue #337, opened 2025-01-03
Cart view OSC pages 


zax71:
**Is your feature request related to a problem? Please describe.**
When running my game show, I’d like to see at  glance that all of the cues are running correctly, so being on the right page for the context of the game would be a nicety 

**Describe the solution you'd like**
OSC messages for next/previous page and set page to `<name>`. These could be configurable with the current cue control system, just right click on the tab and have options for go to this cue and maybe the next/previous in global OSC plugin settings, although, I wouldn’t be using the global ones, just putting this here to flesh out the feature for all use cases.

**Describe alternatives you've considered**
Keypress emulation for the left right arrows would be jank. I could also, just *not* switch pages… But the ability to see at a glance if my OSC setup is working is nice.

**Additional context**
My OSC Controller would be Open Stage Control and I’m changing between 4 pages.

FrancescoCeruti:
I've update the title, the next/previous request overlaps with #255 (which is now possible with the develop version)

Issue #255, opened 2022-08-25
Change page with midi CC


wargreen:
**Is your feature request related to a problem? Please describe.**
When controlling LiSP with a akai MPD 24, i have one page by MPD bank. When changing of bank, i need to use the mouse to change the page onscreen.
Since the MPD don't send MIDI when changing of bank, have the control of the display from the MIDI controller will be very useful for stay focus on the show.

**Describe the solution you'd like**
A way to define a CC (0-127 / page 0 - last page) or note (one by page and/or page -, page +) for changing pages.
For my case i can use only CC, but a logic for everyone seem better !

**Describe alternatives you've considered**
No other way than use the mouse for now.

FrancescoCeruti:
Next/previous page should be relatively easy to build into the "controller" plugin, to allow mapping MIDI or any supported protocol.

As a workaround, since the tabs can be switched with keyboard arrows, you might be able to map the MIDI messages to a keystroke

PR #339, opened 2025-01-03
Navigate to previous/next cart page via OSC/MIDI

s0600204:
This PR adds the ability for users to define a keyboard shortcut, or a MIDI or OSC message that, upon receipt, will cause LiSP to bring up the previous or next page in the Cart Layout.

(It's not possible for the message to specify a specific Page to go to, however - that would require a much larger change.)

Refs. #255, #337


FrancescoCeruti:
I've merged a slightly modified version of this. Thanks :+1:

PR #346, opened 2025-04-06
Time formatting

BryanEnneking:
Here are the changes talked about in discussion #345 

- Durations are now displayed with accurate hundredths of seconds in List Layout
- Pre/post wait times in the cue settings are editable in HH:mm:ss.zzz format
- Media start/stop times are in HH:mm:ss.zzz format
- Spin buttons for pre/post wait and start/stop times manipulate seconds by default

I have spent some time reviewing and testing.  Let me know if you see any problems.

### DIFF
--- a/lisp/core/util.py
+++ b/lisp/core/util.py
@@ -105,19 +105,19 @@ def time_tuple(milliseconds):
     return hours, minutes, seconds, milliseconds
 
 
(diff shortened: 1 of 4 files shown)


FrancescoCeruti:
Thanks :+1:

Issue #86, opened 2017-06-18
Pre/Post wait time displayed as downcounter


xtrm-patrikb:
Hello,

I'll be using your software in a live show for a dancing school and it is just awesome!

I'm using list layout with media cues (to play music files) and action cues (to coordinate participants).
In that context, it would be perfect to display pre/post wait timer as down-counter (time left before next action) more than up-counter.

Many Thanks for that great software!

FrancescoCeruti:
Hi,

Thanks for the feedback, it's a good idea :)

PR #350, opened 2025-04-14
Pre/Post Wait Countdown (Issue #86)

BryanEnneking:
This is a draft to implement the feature requested in Issue #86  After looking at it, I realize there is a much cleaner implementation, so I will update the pull request soon.

The primary thing I want to ask about @FrancescoCeruti is the duration of Media Cues when there is start/stop time specified.  The documentation mentions that duration does not account for start/stop.  Is that intentional, or just because it was not implemented?  The reason I ask is because it's a little strange to count down to a time greater than zero.  If it simply wasn't implemented, I will also take a stab at accounting for start/stop in the duration.  I think it's mostly straight forward (except for seeking to positions outside the start/stop).

### DIFF
--- a/lisp/plugins/list_layout/list_widgets.py
+++ b/lisp/plugins/list_layout/list_widgets.py
@@ -274,6 +274,11 @@ class PreWaitWidget(TimeWidget):
         self.waitTime = CueWaitTime(self.cue, mode=CueWaitTime.Mode.Pre)
         self.waitTime.notify.connect(self._updateTime, Connection.QtQueued)
 
+    def _updateTime(self, time):
+        self.setValue(time)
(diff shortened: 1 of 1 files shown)


FrancescoCeruti:
> The primary thing I want to ask about @FrancescoCeruti is the duration of Media Cues when there is start/stop time specified. The documentation mentions that duration does not account for start/stop. Is that intentional, or just because it was not implemented?

It hasn't been implemented, go for it :rocket: (on a separate branch)

BryanEnneking:
I cleaned it up quite a bit, but let me know if you want it refactored further/differently.

I'll create another branch and pull request for the start/stop duration (probably in the next few days).

Issue #362, opened 2025-07-03
Need Light theme when usin flatpak


Stav88:
**Is your feature request related to a problem? Please describe.**
When using outside  during a sunny day, dark theme make the application very difficulte to read.

**Describe the solution you'd like**
Need a light theme in flatpak container. It Could be QT Fusion theme

**Describe alternatives you've considered**
When runing directly from python, it is possible to configure System theme from preferences window :

![Image](https://github.com/user-attachments/assets/6c2a516b-e6d7-4b50-b0f6-fd0c1a262a35)

But when using flatpak build, the Fusion theme is not installed in the container and there is not other theme usable :

![Image](https://github.com/user-attachments/assets/474b595c-5e56-428d-9540-c925b5435076)

FrancescoCeruti:
There's only a "dark" theme in the codebase ... not sure where those two come from, did you add them?

Stav88:
Sorry, I fogot that I have done a POC from several months by adding some code to have system style in preferences window.

You cans see the code of this POC here : https://github.com/FrancescoCeruti/linux-show-player/compare/master...Stav88:linux-show-player:362-allow-system-theme

Stav88:
I have made a PR to support system themes : https://github.com/FrancescoCeruti/linux-show-player/pull/363

PR #363, opened 2025-07-06
#362 Add current system theme and Support others system theme

Stav88:
New theme combo box :
* In standalone python (Ubuntu 24.04, Gnome)
  image
* In flatpak package
  image

I have add Current sytem theme as user can chose him self. And list all theme installed on system.


FrancescoCeruti:
Some widgets need custom styles to work properly, using random themes might "break" the UI, so this is a no for me.

I've pushed a "light" theme into the master, it's the default "Fusion" style, so a few things will look out of place, you can grab the CI build for that (when it's ready).

I've also started playing around with a proper light theme, and some custom widgets needs to be "style" aware, it will likely come with next "major" release.

PR #365, opened 2025-07-12
Cue Type Icons

BryanEnneking:
I have implemented selectable icons for each cue with defaults for each type of cue.  I couldn't find an icon set that had the consistent look and feel I was going for, so I made my own.  It doesn't have every icon imaginable, but I tried to anticipate all of the broad categories (music, lights, microphones, mechanical, fog, pyrotechnics, etc.).

Originally, I was going to make a new column for cue type icons, but then came up with the idea of combining the status icons and the cue type icons.  In other words, the cue type icons change color based on status.  Let me know what you think.  If you don't like that idea, I can switch back to a separate column.  Also, let me know if there are any cue categories I missed or if there's anything else you don't like.


FrancescoCeruti:
This looks interesting :+1:  mark it as "ready" when you think you're done

BryanEnneking:
I think it's finished.  I just kept it as draft in case you saw changes you wanted right away.

lisp/ui/settings/cue_pages/cue_appearance.py
```diff
@@ -47,6 +57,12 @@ def __init__(self, **kwargs):
         self.cueNameEdit = QLineEdit(self.cueNameGroup)
         self.cueNameGroup.layout().addWidget(self.cueNameEdit)
 
+        # Icon
+        self.iconSelectorDialog = None
+        self.cueIcon = QPushButton("")
```
FrancescoCeruti: At first a glance, it's not very clear what the button is for, we could add a "Change Icon" label, and change the group name to "Cue name and Icon".
This is very noticeable when the "led" (blank) icon is selected.

lisp/ui/settings/cue_pages/cue_appearance.py
```diff
@@ -117,12 +133,65 @@ def enableCheck(self, enabled):
         self.setGroupEnabled(self.fontSizeGroup, enabled)
         self.setGroupEnabled(self.colorGroup, enabled)
 
+    def buildIconSelector(self):
```
FrancescoCeruti: I think we should move this logic to a dedicated QDialog subclass, and simply create the dialog. We can keep in the same file

lisp/ui/settings/cue_pages/cue_appearance.py
```diff
@@ -117,12 +133,65 @@ def enableCheck(self, enabled):
...
+        layout = QVBoxLayout(dialog)
+        grid = QGridLayout()
+        layout.addLayout(grid)
+
+        icon_names = set()
+        path = os.path.join(ICON_THEMES_DIR + "/lisp/cues", "**")
+        for path in glob.iglob(path, recursive=False):
+            if os.path.isfile(path):
+                name, ext = os.path.splitext(os.path.basename(path))
+                icon_names.add(name)
```
FrancescoCeruti: We're looking into a single folder, no reason to use glob, we can use `os.scandir`

lisp/ui/settings/cue_pages/cue_appearance.py
```diff
@@ -117,12 +133,65 @@ def enableCheck(self, enabled):
...
+        icon_names = sorted(icon_names)
+
+        for index, name in enumerate(icon_names):
```
FrancescoCeruti:
```suggestion

        for index, name in enumerate(sorted(icon_names)):
```

lisp/plugins/cart_layout/cue_widget.py
```diff
@@ -316,7 +316,12 @@ def _clicked(self, event):
                     self.cueExecuted.emit(self._cue)
 
     def _updateStyle(self, stylesheet):
-        self.nameButton.setStyleSheet(stylesheet)
+        self.nameButton.setStyleSheet(
+            stylesheet
+            + " background-image: url('lisp/ui/icons/lisp/cues/variations/"
+            + self._cue.icon + "-cart.svg');"
+            + " background-position: center;"
+            + " background-repeat: no-repeat;")
```
FrancescoCeruti: This doesn't scale with the UI. We also should use `ICON_THEMES_DIR`

FrancescoCeruti (changes requested):
I like the result, a few things needs to be addressed before merging.

Other than the code comments: use f-stings to combine variables and strings.

BryanEnneking:
Good feedback.  I'll start working on all of these updates.

lisp/plugins/cart_layout/cue_widget.py
BryanEnneking: I believe I tried background-size and it didn't work.  I'll try some more things.

FrancescoCeruti: I'm not sure if it can be done via stylesheet, but it's a problem only if the window is very small, if you don't find a decent solution it's ok.

s0600204: <details>

<summary>Perhaps something like this?</summary>

```diff
diff --git a/lisp/plugins/cart_layout/cue_widget.py b/lisp/plugins/cart_layout/cue_widget.py
index dfbea630..ee87b9d2 100644
--- a/lisp/plugins/cart_layout/cue_widget.py
+++ b/lisp/plugins/cart_layout/cue_widget.py
@@ -316,12 +316,8 @@ class CueWidget(QWidget):
                     self.cueExecuted.emit(self._cue)
 
     def _updateStyle(self, stylesheet):
-        self.nameButton.setStyleSheet(
-            stylesheet
-            + " background-image: url('lisp/ui/icons/lisp/cues/variations/"
-            + self._cue.icon + "-cart.svg');"
-            + " background-position: center;"
-            + " background-repeat: no-repeat;")
+        self.nameButton.setStyleSheet(stylesheet)
+        self.nameButton.setIcon(IconTheme.get(f'{self._cue.icon}-cart'))
 
     def _enterFadein(self):
         p = self.timeDisplay.palette()
diff --git a/lisp/ui/widgets/qclicklabel.py b/lisp/ui/widgets/qclicklabel.py
index 069209bb..01e5eee1 100644
--- a/lisp/ui/widgets/qclicklabel.py
+++ b/lisp/ui/widgets/qclicklabel.py
@@ -16,6 +16,7 @@
 # along with Linux Show Player.  If not, see <http://www.gnu.org/licenses/>.
 
 from PyQt5.QtCore import pyqtSignal, QEvent
+from PyQt5.QtGui import QIcon, QPainter
 from PyQt5.QtWidgets import QLabel
 
 
@@ -24,7 +25,29 @@ class QClickLabel(QLabel):
 
     def __init(self, parent):
         super().__init__(parent)
+        self._icon = None
+
+    def setIcon(self, icon: QIcon):
+        self._icon = icon
 
     def mouseReleaseEvent(self, e):
         if self.contentsRect().contains(e.pos()):
             self.clicked.emit(e)
+
+    def paintEvent(self, event):
+        if self._icon:
+            size = self.size()
+            w = size.width()
+            h = size.height()
+            b = min(w, h)
+            x = int((w - b) / 2)
+            y = int((h - b) / 2)
+
+            pixmap = self._icon.pixmap(size)
+
+            painter = QPainter()
+            painter.begin(self)
+            painter.drawPixmap(x, y, b, b, pixmap)
+            painter.end()
+
+        super().paintEvent(event)

```
</details>

s0600204:
As an aside, as it might be out-of-scope for this PR: it is theoretically possible to recolour svg icons in Qt programmatically. Could help reduce the number of new files added (and permit colours to be set centrally/follow application theming).

BryanEnneking:
I have committed most of the changes from the comments.  I'll work on the Cart Layout resizing next.

BryanEnneking:
Everything should be ready now.  The Cart Layout icons now scale appropriately.  I may try recoloring the icons as suggested, but that will be for another PR.

FrancescoCeruti:
I did some refactoring to make the dialog code a bit easier, using a QListView, it can also be navigated by keyboard.

Thanks :rocket:

PR #367, opened 2025-08-09
Re-color Cue Type Icons

BryanEnneking:
Instead of having separate files for each variation of a cue type icon (running, pause, error, cart), this code re-colors the base icons at load time.  The result is a more maintainable icon set.  To make this possible, the icons needed a little SVG refactoring so color and opacity could be changed in one place.  If the implementation isn't what you were expecting, I'm open to making changes.


s0600204:
Modifying the `.svg` attributes programatically. An interesting approach, and one I had not considered. I wonder if that would work with the other icons used within the UI of LiSP. (In case we ever want to support recolouring the UI icons based on the active theme, or for users with some variety of colour-blindness.)

https://github.com/s0600204/linux-show-player/commit/38be8fdc1e4e2262bdce82397141bac405fadf27 was my take on it.

BryanEnneking:
Yes.  I think that could be possible.  The other icons might need a little refactoring though.

Issue #322, opened 2024-07-04
Add a converter / importer for older lsp files


pappou99:
I have a bunch of showfiles, which I'm not able to open with the newer Versions of LISP.(< 0.6)

Please add a possibility to import older files, so LISP is backwards compatible or provide informations how to convert these older files.

FrancescoCeruti:
There's been a few discussion about this, with 0.6 LiSP saves the app and plugin version in the session file, the next step is to  convert the file automatically to newer version (with a notice for the user).

I might be able to provide 0.5 -> X conversion when it's added

PR #369, opened 2025-09-24
Session file converter

BryanEnneking:
Here is a converter to automatically detect and upgrade the session file from version 0.5 to the current version.  I took a minimalist approach by only making the necessary changes to make the session open and usable without errors.  I did my best to craft test sessions in 0.5 and load them in 0.6.5.  Most likely there are edge cases that got missed, so it would be nice for others to test with real-world sessions.

### DIFF
--- a/lisp/application.py
+++ b/lisp/application.py
@@ -28,6 +28,7 @@ from lisp.core.session import Session
 from lisp.core.signal import Signal
 from lisp.core.singleton import Singleton
 from lisp.core.util import filter_live_properties
+from lisp.core.version_manager import VersionManager
 from lisp.cues.cue import Cue
 from lisp.cues.cue_factory import CueFactory
 from lisp.cues.cue_model import CueModel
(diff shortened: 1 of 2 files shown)


BryanEnneking:
This addresses issues such as #322 and others.

s0600204:
I'm not sure you're aware, but in case you aren't: there exists some work on session conversion done already - found on Francesco's [`session_migrations`](https://github.com/FrancescoCeruti/linux-show-player/tree/session_migrations) branch ([diff](https://github.com/FrancescoCeruti/linux-show-player/compare/develop...session_migrations))

A cursory glance shows some differences and some similarities. Not sure if one approach is better than the other, or if it is possible to consolidate both your work and his into one.

BryanEnneking:
I was not aware, but thank you for letting me know.  That is a very different approach.  Feel free to use my code if it's useful or get rid of it if it's not.

FrancescoCeruti:
Thanks! I've integrated some of your changes into #377

Issue #378, opened 2026-01-06
Defining Wayland's app_id


tkna91:
Currently, the Linux Show Player V0.6.5 does not appear to have a Wayland app_id.
Having an app_id defined would be useful because it makes it easier to automate window management.
For example, it would allow actions like bringing focus back to the Linux Show Player after focusing on an external window.
It is still possible to do this using window titles, but accidents may occur when similar window titles exist.
Being able to override it via a command-line option like `--app-id=APPID` might be preferable, as it would simplify managing multiple instances.

<img width="1037" height="783" alt="Image" src="https://github.com/user-attachments/assets/069e95eb-5264-4386-ac59-9ee31e873446" />

PR #396, opened 2026-09-07
fix: set Wayland desktop file name

keeltrace:
AI-generated contribution. I am the human operator responsible for follow-up with maintainers. Please feel free to close this PR or request any changes.

Addresses #378 with the minimal stable app ID setup.

Qt derives the Wayland app_id from the desktop file name when it is set. Linux Show Player already ships `dist/linuxshowplayer.desktop`, so set the QApplication desktop file name to `linuxshowplayer` during startup.

This keeps the existing application name unchanged and does not add the optional CLI override proposed in the issue.

Validation:
- `git diff --check`
- `python3 -m py_compile lisp/main.py`

PyQt5 is not installed in the local validation environment, so no GUI runtime smoke was performed.

### DIFF
--- a/lisp/main.py
+++ b/lisp/main.py
@@ -108,6 +108,7 @@ def main():
     # Create the QApplication
     qt_app = QApplication(sys.argv)
     qt_app.setApplicationName("Linux Show Player")
+    qt_app.setDesktopFileName("linuxshowplayer")
     qt_app.setQuitOnLastWindowClosed(True)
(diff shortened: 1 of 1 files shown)


FrancescoCeruti:
The flatpak `.desktop` file has a different name, I guess this change will not work in that case? :thinking:

keeltrace:
Good catch, updated

